PR #6, opened 2020-07-13
change user-facing strings from "json" to "JSON" 

toloudis:
All user facing occurrences of the word json will be JSON.  The exception is where the file dialog shows the file extension filter (`.json`) -- file extensions are lowercased by convention.

### DIFF
--- a/agave_app/agaveGui.cpp
+++ b/agave_app/agaveGui.cpp
@@ -318,7 +318,7 @@ agaveGui::saveJson()
 #ifdef __linux__
   options |= QFileDialog::DontUseNativeDialog;
 #endif
-  QString file = QFileDialog::getSaveFileName(this, tr("Save JSON"), QString(), tr("json (*.json)"), nullptr, options);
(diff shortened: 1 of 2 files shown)


heeler:
The json -> JSON looks fine. 

I'm not very familiar with QT but I was interested by all strings being wrapped with tr, ie `tr("blahblabblah")`. 
When I looked it up it stated this was for internationalization/language translation but that you have to implement that in with QtLinguist. Is that something we're doing? If not is the tr( superfluous?

toloudis:
> The json -> JSON looks fine.
> 
> I'm not very familiar with QT but I was interested by all strings being wrapped with tr, ie `tr("blahblabblah")`.
> When I looked it up it stated this was for internationalization/language translation but that you have to implement that in with QtLinguist. Is that something we're doing? If not is the tr( superfluous?

https://doc.qt.io/qt-5/i18n-source-translation.html#using-tr-to-obtain-a-translation

Basically this is the approved "best practice" for Qt user interfaces:  all user-facing strings should be QStrings and all user-facing string literals should be wrapped with tr().  Just doing that now saves potential headaches later if we really want to internationalize it properly.

PR #9, opened 2020-10-07
Feature/docs

toloudis:
Automated end-user documentation generation using Sphinx.

The docs are generated via github-actions in both html and pdf form.
They can also be generated via command line using standard Sphinx "make" commands.
Preliminary setup to publish html docs to gh-pages at allen-cell-animated.github.io/agave .
Pdf docs are included in build artifacts alongside executables.

There are a few minor warning fixups and versioning/build environment updates that came along for the ride in this PR.


.github/workflows/tagged_master_release.yml
```diff
@@ -102,6 +99,35 @@ jobs:
...
+      - uses: actions/upload-artifact@v1
+        with:
+          name: DocumentationPDF
+          path: docs/_build/latex/agave.pdf
+      - name: Publish Docs
+        uses: JamesIves/github-pages-deploy-action@releases/v3
+        with:
+          ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
+          BASE_BRANCH: master # The branch the action should deploy from.
+          BRANCH: gh-pages # The branch the action should deploy to.
```
EricJIsaac, line 129: Is this the site you are expecting the workflow to deploy the docs to?
https://allen-cell-animated.github.io/agave/

agave_app/renderer.cpp
```diff
@@ -303,7 +303,6 @@ Renderer::reset(int from)
   glEnable(GL_LINE_SMOOTH);
 
   this->m_time.start();
-  this->m_time = this->m_time.addMSecs(-from);
```
EricJIsaac, line 306: Any reason not to remove the unused parameter `from` from the Render::reset function?

EricJIsaac (approved):

.github/workflows/tagged_master_release.yml
JacksonMaxfield: That should be the correct link yep!

JacksonMaxfield:
@toloudis @heeler GitHub added ["org level" secrets](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-organization)

I just set them up for my own civic stuff but you may want to set them up for this org and allencellmodeling. It greatly reduces the amount of various keys we have lying around.

i.e. instead of each repo having a `ACCESS_TOKEN` secret, you have a single `ACCESS_TOKEN` secret for the entire org. It also makes setting up new repos _much_ simpler because, in the case of the python cookiecutter, a new repo would only need to add the `CODECOV_TOKEN`, as the `PYPI_TOKEN` and the `ACCESS_TOKEN` can be org level.

renderlib/CMakeLists.txt
```diff
@@ -71,15 +71,15 @@ add_compile_definitions(_STATICLIBBUILD)
 # I prefer target_compile_definitions but the target has to exist first
 add_subdirectory(libCZI/Src/JxrDecode)
 add_subdirectory(libCZI/Src/libCZI)
-find_library(libCZI REQUIRED PATHS ${CMAKE_SOURCE_DIR}/libCZI)
+find_library(libCZI PATHS ${CMAKE_BINARY_DIR}/renderlib/libCZI/Src/libCZI)
```
heeler, line 74: What's the reasoning for removing the REQUIRED keyword?

heeler (approved):
A few minor questions but nothing seems wrong. Nice job automating this.

.github/workflows/tagged_master_release.yml
toloudis: Yes.  Fingers crossed.

agave_app/renderer.cpp
toloudis: The only reason is to minimize the changes here.  As I recall there was some warning coming from Qt about addMSecs and this was just cleaning it up.  It's really a minor change but you're right to think about whether the from time will ever be needed at all.

toloudis: Looking again, I believe it was that QTime is deprecated in some way and QElapsedTimer is the more appropriate class to use for this kind of measurement.

CMake/Info.plist.cmake
```diff
@@ -19,9 +19,9 @@
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>@PACKAGE_VERSION@</string>
+	<string>@PROJECT_VERSION@</string>
```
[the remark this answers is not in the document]
toloudis: Yes this is a template for the true macos Info.plist.  CMake can fill it in with cmake variables inside the `@` symbols.  The goal here is to put the "version" in only one place, and for now that one place is the top of the master CMakeLists.txt file.  There's a cmake step that turns this file into its final form as Info.plist.

renderlib/CMakeLists.txt
toloudis: To be honest, that may be an accidental removal for a scenario where it wasn't finding it in the right place.  It's clearly required implicitly anyway, as the build will fail without it.

toloudis:
> i.e. instead of each repo having a `ACCESS_TOKEN` secret, you have a single `ACCESS_TOKEN` secret for the entire org. It also makes setting up new repos _much_ simpler because, in the case of the python cookiecutter, a new repo would only need to add the `CODECOV_TOKEN`, as the `PYPI_TOKEN` and the `ACCESS_TOKEN` can be org level.

Just curious, why is the CODECOV_TOKEN different?

.github/workflows/tagged_master_release.yml
EricJIsaac: Ok! Should this be linked in the README or is it generally understood on GitHub that this is where a dev should look for docs?

JacksonMaxfield:
> Just curious, why is the CODECOV_TOKEN different?

The codecov token is provided by codecov. It is the auth to upload the codecov.xml file to their service to track code coverage over time. Because it is tracking a single repos code coverage, the token provided by codecov is linked to the repo.

Thats just how they have it setup on their side. but a feature request to them allowing org level tokens would be great. I wonder if they are working on it....

.github/workflows/tagged_master_release.yml
JacksonMaxfield: Seconding adding a link to the README.

toloudis: This is not dev documentation but end-user documentation.  The README is really not for end-user documentation in this case.  It's worth adding a link there, sure, but really we'd want to link it from allencell.org and maybe put a link in the app itself as well.  I'm bundling the same documentation as a PDF with the app so end users will be able to see it that way too.

JacksonMaxfield:
Following up on this.... Maybe the codecov token isn't needed for many of our repos: https://github.com/codecov/codecov-action#usage

Codecov can just work on public repos it seems. Which makes sense.

PR #12, opened 2021-01-26
Feature/load float

toloudis:
Allow loading TIFF files with single-sample float32 pixels.  RGB(A) tiff still not supported.

The basic idea here is that the native internal volume format for Agave is uint16.  In order to load float32, I load a whole channel, retain the min and max, and rescale the float values to fill the full uint16 range (min->0 to max->65535).

With this change, Agave can now open tiff files with scalar uint8, uint16, and float32 pixels.

### DIFF
--- a/renderlib/FileReaderTIFF.cpp
+++ b/renderlib/FileReaderTIFF.cpp
@@ -106,8 +106,12 @@ readTiffDimensions(TIFF* tiff, const std::string filepath, VolumeDimensions& dim
 
   uint16_t sampleFormat = SAMPLEFORMAT_UINT;
   if (TIFFGetField(tiff, TIFFTAG_SAMPLEFORMAT, &sampleFormat) != 1) {
-    // just warn here.  We are not yet using sampleFormat!
-    LOG_WARNING << "Failed to read sampleformat of TIFF: '" << filepath << "'";
+    LOG_ERROR << "Failed to read sampleformat of TIFF: '" << filepath << "'";
+    return false;
+  }
+  if (sampleFormat != SAMPLEFORMAT_UINT && sampleFormat != SAMPLEFORMAT_IEEEFP) {
+    LOG_ERROR << "Unsupported tiff SAMPLEFORMAT " << sampleFormat << " for '" << filepath << "'";
+    return false;
   }
 
   uint32_t sizeT = 1;
@@ -244,8 +248,8 @@ readTiffDimensions(TIFF* tiff, const std::string filepath, VolumeDimensions& dim
     QString pixelType = pixelsEl.attribute("Type", "uint16").toLower();
     LOG_INFO << "pixel type: " << pixelType.toStdString();
     bpp = mapPixelTypeBPP[pixelType.toStdString()];
-    if (bpp != 16 && bpp != 8) {
-      LOG_ERROR << "Image must be 8 or 16-bit integer typed";
+    if (bpp != 32 && bpp != 16 && bpp != 8) {
+      LOG_ERROR << "Image must be 8 or 16-bit integer, or 32-bit float typed";
       return false;
     }
     sizeX = requireUint32Attr(pixelsEl, "SizeX", 0);
@@ -309,6 +313,7 @@ readTiffDimensions(TIFF* tiff, const std::string filepath, VolumeDimensions& dim
   dims.physicalSizeZ = physicalSizeZ;
   dims.bitsPerPixel = bpp;
   dims.channelNames = channelNames;
+  dims.sampleFormat = sampleFormat;
 
   dims.log();
 
@@ -332,6 +337,19 @@ copyAndConvert(uint8_t* dest, const uint8_t* src, size_t numBytes, int srcBitsPe
       dataptr16++;
     }
     return numBytes * 2;
+  } else if (srcBitsPerPixel == 32) {
+    // convert pixels
+    // this assumes tight packing of pixels in both buf(source) and dataptr(dest)
+    uint16_t* dataptr16 = reinterpret_cast<uint16_t*>(dest);
+    const float* src32 = reinterpret_cast<const float*>(src);
+    for (size_t b = 0; b < numBytes / 4; ++b) {
+      // BIG ASSUMPTION: float data is in 0..1 range!
+      // I am not computing a global max and rescaling to that max.
+      // This operation will overflow/underflow for anything outside 0..1
+      *dataptr16 = (uint16_t)(src32[b] * 65535.0);
+      dataptr16++;
+    }
+    return numBytes / 2;
   } else {
     LOG_ERROR << "Unexpected tiff pixel size " << srcBitsPerPixel << " bits";
     return 0;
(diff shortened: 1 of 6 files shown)


renderlib/FileReaderTIFF.cpp
```diff
@@ -317,21 +321,56 @@ readTiffDimensions(TIFF* tiff, const std::string filepath, VolumeDimensions& dim
...
 {
+  // how many pixels in this channel:
+  size_t numPixels = dims.sizeX * dims.sizeY * dims.sizeZ;
+  int srcBitsPerPixel = dims.bitsPerPixel;
+
   // dest bits per pixel is IN_MEMORY_BPP which is currently 16, or 2 bytes
   if (IN_MEMORY_BPP == srcBitsPerPixel) {
-    memcpy(dest, src, numBytes);
-    return numBytes;
+    memcpy(dest, src, numPixels * (srcBitsPerPixel / 8));
```
heeler, line 343: it's clear what you are doing but I'd still prefer to see numPixels * srcBytesPerPixel rather to make it crystal clear. (it's a little pedantic so go with your judgement)

renderlib/FileReaderTIFF.cpp
```diff
@@ -317,21 +321,56 @@ readTiffDimensions(TIFF* tiff, const std::string filepath, VolumeDimensions& dim
...
+    return 1;
   } else if (srcBitsPerPixel == 8) {
-    // convert pixels
-    // this assumes tight packing of pixels in both buf(source) and dataptr(dest)
     uint16_t* dataptr16 = reinterpret_cast<uint16_t*>(dest);
-    for (size_t b = 0; b < numBytes; ++b) {
+    for (size_t b = 0; b < numPixels; ++b) {
```
heeler, line 347: I'm assuming you're using a uint16 to copy 2 at a time? 
What if you have an odd number of 8-bit pixels? 
Why are you storing an 8-bit pixel in a uint16?

renderlib/FileReaderTIFF.cpp
```diff
@@ -317,21 +321,56 @@ readTiffDimensions(TIFF* tiff, const std::string filepath, VolumeDimensions& dim
...
       *dataptr16 = (uint16_t)src[b];
       dataptr16++;
     }
-    return numBytes * 2;
+    return 1;
+  } else if (srcBitsPerPixel == 32) {
+    // assumes 32-bit floating point (not int or uint)
+    uint16_t* dataptr16 = reinterpret_cast<uint16_t*>(dest);
```
heeler, line 354: Why are you compressing your numerical image range? 
Is this to save space or ? 

I suspect there's a reason you are using uint16 throughout your types but it's not clear why, maybe add a comment near here somewhere?

renderlib/FileReaderTIFF.cpp
```diff
@@ -372,7 +411,7 @@ readTiffPlane(TIFF* tiff, int planeIndex, const VolumeDimensions& dims, uint8_t*
       return false;
     }
     // copy buf into data.
-    size_t numBytesCopied = copyAndConvert(dataPtr, reinterpret_cast<uint8_t*>(buf), numBytesRead, dims.bitsPerPixel);
+    size_t numBytesCopied = copyDirect(dataPtr, reinterpret_cast<uint8_t*>(buf), numBytesRead, dims.bitsPerPixel);
```
heeler, line 414: You are using reinterpret_cast a lot. I looked it up and am a little concerned that the consensus is it makes your code non-portable. Can you comment on this?

renderlib/FileReaderTIFF.cpp
```diff
@@ -496,15 +540,30 @@ FileReaderTIFF::loadOMETiff(const std::string& filepath, VolumeDimensions* outDi
...
   uint8_t* destptr = data;
 
+  // still assuming 1 sample per pixel (scalar data) here.
+  size_t rawPlanesize = dims.sizeX * dims.sizeY * (dims.bitsPerPixel / 8);
+  // allocate temp data for one channel
+  uint8_t* channelRawMem = new uint8_t[dims.sizeZ * rawPlanesize];
+  memset(channelRawMem, 0, dims.sizeZ * rawPlanesize);
+
+  // stash it here in case of early exit, it will be deleted
+  std::unique_ptr<uint8_t[]> smartPtrTemp(channelRawMem);
```
heeler, line 550: I'd use a container class here. std::array

renderlib/VolumeDimensions.h
```diff
@@ -14,6 +14,10 @@ struct VolumeDimensions
   float physicalSizeY = 1.0f;
   float physicalSizeZ = 1.0f;
   uint32_t bitsPerPixel = 16;
+  // SAMPLEFORMAT_UINT = 1;
```
heeler, line 17: What's this about?

heeler (approved):
My comments here are minor. I'm a big fan of STL and honestly, it freaks me out a bit to see raw indexes. The only other points are more comments and a bit of concern over your abundant use of reinterpret_cast which I'm not so familiar with but have read that it can have some unpleasant consequences.

renderlib/FileReaderTIFF.cpp
```diff
@@ -317,21 +321,56 @@ readTiffDimensions(TIFF* tiff, const std::string filepath, VolumeDimensions& dim
 
 // return number of bytes copied to dest
 size_t
-copyAndConvert(uint8_t* dest, const uint8_t* src, size_t numBytes, int srcBitsPerPixel)
+copyDirect(uint8_t* dest, const uint8_t* src, size_t numBytes, int srcBitsPerPixel)
```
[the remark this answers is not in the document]
toloudis: The function used to be nontrivial.  I left it as a function mostly to minimize the amount of changes, and allow for the implementation to be modified later, keeping it encapsulated to this function.

renderlib/FileReaderTIFF.cpp
toloudis: This function is converting everything to 16-bit uints.  This code branch is converting 8-bit uints (src) to 16-bit uints (dest).

renderlib/FileReaderTIFF.cpp
toloudis: This function is converting everything to 16-bit uints. This code branch is converting 32-bit floats (src) to 16-bit uints (dest).

renderlib/FileReaderTIFF.cpp
toloudis: hmm, thanks - I think you're right to flag this.  Going from void* to uint8_t* is totally reasonable for static_cast.  I will make that change.

renderlib/VolumeDimensions.h
toloudis: sampleFormat basically keeps track of whether the original source data was float or uint

toloudis: in this case I am using the libtiff-specific values, so if I make changes for reading CZI files I'll have to reconcile things at that time.

renderlib/FileReaderTIFF.cpp
toloudis: great suggestion, thanks!

EricJIsaac (approved):

PR #13, opened 2021-04-30
1.1.0 updates for time series data

toloudis:
This is a multi-PR with a few different changes.  I apologize for my laziness in not splitting this up a bit more but the changes are fairly targeted and discernible when you look.  This work happened mostly when trying to render a long time sequence from the nuclear morphogenesis data set.

This should result in a agave 1.1.0 version release after merged. 

* update the Qt dependency version to 5.15.2  
* recommend using the python library aqtinstall to help with automatic Qt install

The remaining changes are mostly around supporting rendering long time lapse sequences.

* several bugfixes in which current scene and / or current time was not kept up to date
* when changing time, update the transfer function levels such that the absolute intensity ranges are preserved (this is implemented in the GradientData::convert function in GradientData.cpp, and called from TimelineDockWidget and from the SetTimeCommand::execute).  This change is for consistency on the assumption that it is more desirable to have absolute intensity thresholding from one time sample to the next.  Results are good for the one long time sequence render I did.
* add a SetTime command for the python and the websocket interfaces
* increase the number of histogram bins to 512, giving more precision when dialing in thresholds using the gradient editor


agave_app/python/pyrenderer.cpp
```diff
@@ -196,12 +196,20 @@ OffscreenRenderer::LoadOmeTif(const std::string& s)
 }
 // load a volume
 int
-OffscreenRenderer::LoadVolumeFromFile(const std::string& s, int time, int scene)
+OffscreenRenderer::LoadVolumeFromFile(const std::string& s, int scene, int time)
```
kmitcham, line 199: Changing the ordering of a pair of like-typed args is always worrisome- can create hard to find bugs.

kmitcham (approved):

agave_app/python/pyrenderer.cpp
toloudis: or in this case, fix them 😟

vianamp (approved):
Looking forward to try this out.

PR #14, opened 2021-05-31
Feature/aws egl

toloudis:
Several code changes to enable agave to run as a websocket server on AWS linux:
* on Linux, use EGL to initialize OpenGL for headless server mode operation, rather than standard Qt OpenGL (which just doesn't work at all).
* remove some of the Qt dependency OpenGL wrapper classes (QOpenGLFramebuffer and QOpenGLShaderProgram) and replace with simplified wrapper classes that don't require Qt
* dockerize the app for headless server mode, using Qt's `-platform offscreen` plugin
* add a terraform script to spin up the (hopefully) minimal infrastructure to simply push the docker image to deploy
* simplify the pythonclient python module to easily connect to a local agave (run with `agave --server`) and render locally via websockets  
* bug fix for a couple of commands that were not being parsed by the websocket server


agave_app/renderer.cpp
```diff
@@ -64,11 +62,11 @@ Renderer::init()
   // QMessageBox::information(this, "Info:", "Application Directory: " + QApplication::applicationDirPath() + "\n" +
   // "Working Directory: " + QDir::currentPath());
 
-  QSurfaceFormat format = renderlib::getQSurfaceFormat();
-
-  this->m_glContext = new QOpenGLContext();
-  this->m_glContext->setFormat(format); // ...and set the format on the context too
-  this->m_glContext->create();
+#if HAS_EGL
```
GabeMedrash, line 65: Just for my understanding: you're using compiler directives(?)--here and elsewhere in this diff--to ensure that only certain branches of code get compiled on different platforms?

toloudis: That's correct.  The HAS_EGL flag only gets defined if compilation is happening on Linux.   All those `#if`s get preprocessed before compile time and as a result only the HAS_EGL side gets compiled into the linux build.

toloudis:
I forgot to mention this but I would love feedback on the terraform script in particular.  That's my first ever attempt with terraform and I wasn't sure about best practices, whether to keep it in the same repo as the app, etc.

ci/main.tf
```diff
@@ -0,0 +1,242 @@
+#########################################
```
Aerendel, line 1: I see some high level things you'll want to verify / dive deeper into with @anl , @chris-esterlein and / or @griffinfujioka :

- I don't see a backend defined which I imagine means currently will only use a local backend to manage the terraform state (ie. your laptop). Could be wrong but I believe you'll need to define a remote backend (SW team uses GCS as a backend I believe)
- The AWS "default" VPC should really never be used. You want to be using one of the custom VPCs that we have setup. This configuration will be different whether you're deploying to the Dev account or the Prod account.
- Not sure where you've been testing this, but as mentioned above, we have a Dev and a Prod account for which some configurations may vary: specifically VPC / AZ and potentially security group settings. To manage different environment targets effectively, you'll want to introduce terraform var files per environment
- I'm personally a fan of having the terraform code in the same repo, however the current pattern we have is setup with a monorepo (on bitbucket for now) -> https://aicsbitbucket.corp.alleninstitute.org/projects/SW/repos/terraform-platform/browse. The repo has an automatic plan and deploy tool (Atlantis) that has all the proper credentials and configs. I believe this is the pattern we're still expected to use until we come up with something better - once again, confirm with @anl and the other devops folks.

toloudis: Thanks, this is great info.  

We don't currently have plans to deploy to prod or keep this live, for now.  It's really just meant to be up for some short bursts, demo, testing/verification at this time.  Mostly a PoC.

I've been running it in the "AICS Animated Cell" aws account (which probably should go away at some point).   So probably to get one of the custom VPCs I'll have to use my credentials on the SW account instead.  

I'm not familiar with backends in terraform so I'll have to look that one up.

kmitcham (approved):

PR #16, opened 2021-06-28
Feature/python module

toloudis:
The big idea here is to create a python client library for agave.  It does not completely embed agave;  it communicates with a separately running agave server process.

You run `agave --server` in the background and it will wait for connections from the `agave_pyclient`.  Intrepid Python programmers could spawn agave and shut it down within their python scripts, if desired!

Agave GUI also has an "export Python script" command that captures all the current settings in a python script that is compatible with this mode.   So you can run an interactive gui session, capture the settings, and then modify that script to animate them in a loop, etc.

There are also methods in the pyclient to automate a turntable (360 degree rotation around the vertical axis).

The previous python client code is refactored into a new agave_pyclient directory with a whole bunch of cookiecutter boilerplate files.   The pre-existing github actions are modified to add a pypi publish step.

The key python client files are `agave_pyclient/agave_pyclient/agave.py` and `.agave_pyclient/agave_pyclient/commandbuffer.py`.  

It is important to publish both agave and the pyclient in lockstep because the binary communication protocol is very specific and creates tight coupling between server and client.  It could change from version to version.  So for now, it is just way simpler to always publish both.  (A TODO before the next actual 1.1 release is to keep the client and server version numbers the same, using the git tag / CMakeLists version as the source of truth)


.github/workflows/tagged_master_release.yml
```diff
@@ -99,6 +107,33 @@ jobs:
...
+      - name: Install Dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel
+      - name: Build Package
+        run: |
+          cd agave_pyclient
+          python setup.py sdist bdist_wheel
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@master
```
anl, line 130: For third-party actions, GitHub recommends pinning to a SHA:

https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions

(and by its nature, this action gets passed sensitive information).

toloudis: Good idea, will at least pin to a known release tag

PR #17, opened 2021-07-15
Feature/factor out qt

toloudis:
The `renderlib` subdirectory has a lot of Qt dependencies.
That directory is treated within this project as a separate library.  
In order to make the code more reusable (and in part preparing for a next-gen Agave 2.0) I have begin to remove the Qt dependency from renderlib.

This removes most of it, with the last nontrivial exception being the OpenGL initialization.


agave_app/StatisticsWidget.h
```diff
@@ -45,4 +47,33 @@ public slots:
 
 private:
   QGridLayout m_MainLayout;
+
+  class MyStatusObserver : public IStatusObserver
```
toloudis, line 51: This internal class is kind of a silly implementation and I imagine I will redo it with lambda functions to make it more readable and less convoluted, but for now it's working.

toloudis: (It is replacing something that used to use Qt's signals and slots between renderlib and the main app GUI.)

PR #19, opened 2021-08-06
Feature/save python script

toloudis:
1. Reinstate the "Save to Python script" command

2. Bug fixes:
* Most important: fix calls to `stat` (show stopper for python usage)
* Fix Qt linker warnings on macos by incrementing the target deployment version from 10.12 to 10.13
* SetVoxelScale command was changing too many other settings.

### DIFF
--- a/renderlib/command.cpp
+++ b/renderlib/command.cpp
@@ -12,6 +12,7 @@
 #include <QJsonDocument>
 #include <QJsonObject>
 
+#include <errno.h>
 #include <sys/stat.h>
 
 void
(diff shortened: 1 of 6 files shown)


renderlib/command.cpp
```diff
@@ -31,7 +32,8 @@ LoadOmeTifCommand::execute(ExecutionContext* c)
 {
   LOG_WARNING << "LoadOmeTif command is deprecated. Prefer LoadVolumeFromFile command.";
   LOG_DEBUG << "LoadOmeTif command: " << m_data.m_name;
-  if (stat(m_data.m_name.c_str(), nullptr) == 0) {
```
EricJIsaac, line 34: FMI, was stat encountering an error when called w/ nullptr from python but not in cpp? (and if so, do you know why)

EricJIsaac (approved):

renderlib/command.cpp
toloudis: Good question.  I honestly don't know how I was able to load files in GUI but not Python (agave running in server mode).  I haven't tried to repro on any OS other than MacOS. At first I thought it might be a permission issue.  However, the error I was getting indicated that it couldn't handle passing a null buffer in as the second argument.  So I feel much more comfortable that I am calling stat() in a way that should work more universally.  

I'm going to continue with more testing both GUI and server modes on other platforms to make sure these basic functions are working.

toloudis: Oh wait, I do know the answer to the first part.  The stat calls were added as part of an effort to factor Qt out of the "back end".  The front end actually has a similar stat call but uses a higher level Qt function which continues to work when in GUI mode.  So when in gui mode, it doesn't use quite the same function to test file existence.

toloudis: https://github.com/allen-cell-animated/agave/blob/main/agave_app/agaveGui.cpp#L345

PR #20, opened 2021-09-04
Bugfix/cstatus crash on change render

toloudis:
Changing renderer was causing a crash.

Agave has two different rendering modes and each mode is captured in a different IRenderWindow subclass.

The CStatus object, by virtue of being owned by the Renderer, was not surviving a Renderer change, but because it is being shared up at the StatsWidget level, the Stats Widget was holding on to a bad pointer. 

The simple answer is to use shared_ptr to have a refcounted instance of the CStatus object.
An alternative future implementation is suggested in a comment.

### DIFF
--- a/renderlib/IRenderWindow.h
+++ b/renderlib/IRenderWindow.h
@@ -20,7 +20,9 @@ public:
   virtual void resize(uint32_t w, uint32_t h, float devicePixelRatio = 1.0f) = 0;
   virtual void cleanUpResources() {}
 
-  // an interface for reporting statistics and other data updates
+  // An interface for reporting statistics and other data updates
+  // The IRenderWindow is the first to create this object but it can get shared with other widgets in the gui.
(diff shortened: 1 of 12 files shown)


renderlib/IRenderWindow.h
```diff
@@ -18,8 +20,10 @@ class IRenderWindow
   virtual void resize(uint32_t w, uint32_t h, float devicePixelRatio = 1.0f) = 0;
   virtual void cleanUpResources() {}
 
-  // an interface for reporting statistics and other data updates
-  virtual CStatus* getStatusInterface() { return nullptr; }
+  // An interface for reporting statistics and other data updates
+  // The IRenderWindow is the first to create this object but it can get shared with other widgets in the gui.
+  // Alternative impl could be to create at app layer and pass down into renderers
```
kmitcham, line 25: This is a good comment; a nit might be to explain why you used one solution over the alternate.

kmitcham (approved):

PR #24, opened 2021-09-30
update docs

toloudis:
Small update to documentation for the currently available command line arguments

### DIFF
--- a/docs/agave.rst
+++ b/docs/agave.rst
@@ -470,6 +470,31 @@ For Python users, the AGAVE Python client project full documentation can be foun
 https://allen-cell-animated.github.io/agave/agave_pyclient
 
 
+Command Line Interface
+----------------------
+
+AGAVE supports the following command line options:
+
+``--server``
+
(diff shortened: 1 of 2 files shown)


docs/agave.rst
```diff
@@ -470,6 +470,31 @@ For Python users, the AGAVE Python client project full documentation can be foun
...
+
+``--list_devices``
+
+  Only valid in server mode on Linux.  AGAVE will dump a list of possible GPU devices and then exit. 
+
+``--gpu number``
+
+  Only valid in server mode on Linux. Selects a device to use from the list provided by list_devices. The device is specified as a zero-based index into the list.
+
+``-platform offscreen``
```
schoinh, line 494: Does this also need to start with `--` like the others?

toloudis: I know that looks odd but it really IS supposed to only have single hyphen.  It's a built-in Qt feature that tells Qt not to tie in to the native windowing system.

EricJIsaac (approved):

MatteBailey (approved):

PR #29, opened 2022-02-25
fix background color when Save Image happens

toloudis:
When Save Image button was clicked, the background color was not being respected.  Just an oversight, this hadn't been tested.  Simple fix.


kmitcham (approved):
Famous last words, but the change does look safe.

PR #30, opened 2022-03-08
Fix/time series stat warning

toloudis:
Large files were failing to run `stat` on Windows.  This is because plain `stat` can not handle 64 bit file sizes/offsets and you have to use `stat64`.  However, on Apple, plain `stat` is already 64-bit ready and stat64 is deprecated.  

The effect of this bug is that some Python scripts that were trying to load large tiff files were completely failing because the code thinks the stat failure was some indicator of a problem with the file.

### DIFF
--- a/renderlib/command.cpp
+++ b/renderlib/command.cpp
@@ -13,6 +13,20 @@
 #include <errno.h>
 #include <sys/stat.h>
 
+#if defined(WIN32)
+#define STAT64_STRUCT __stat64
+#define STAT64_FUNCTION _stat64
+#elif defined(__APPLE__)
+#define STAT64_STRUCT stat
+#define STAT64_FUNCTION stat
+#elif defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
+#define STAT64_STRUCT stat
+#define STAT64_FUNCTION stat
+#else
+#define STAT64_STRUCT stat64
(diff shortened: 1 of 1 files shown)


renderlib/command.cpp
```diff
@@ -13,6 +13,20 @@
 #include <errno.h>
 #include <sys/stat.h>
 
+#if defined(WIN32)
+#define STAT64_STRUCT __stat64
+#define STAT64_FUNCTION _stat64
+#elif defined(__APPLE__)
+#define STAT64_STRUCT stat
+#define STAT64_FUNCTION stat
+#elif defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
```
kmitcham, line 22: this is not combined with line 19 to avoid getting Apple cooties on the linux code?  Less comically, so that if other os-dependent stuff needs to happen, you can re-use these blocks.

kmitcham (approved):

renderlib/command.cpp
toloudis: I like keeping the apple and linux sides separate even if they have the same result.  If this kind of thing has to happen more often, it's going to all get moved into some kind of a platform.h file and then reshuffled.  Currently I don't support building on any linux besides Ubuntu, but the stuff I found about 64-bit stat made this distinction clear

griffinfujioka (approved):

PR #31, opened 2022-03-10
Feature/better turntable

toloudis:
The method of turntable rotation (a Python interface function) was not accurate to what was intended.  The turntable function is supposed to rotate the current view about the vertical screen axis.   It had been using "orbit" instead of "trackball".  Orbit is using the transformed axes of the original volume as the rotation axes, and so if the volume is rotated off-axis, orbit doesn't have the expected behavior.  Trackball is the correct intended way of rotating for this turntable mode.  It always uses the vertical screen axis regardless of any previous camera positioning.   

Trackball is added as a Python command and the turntable and rocker functions are changed to use it.  There are also a few formatting changes and a small change to the readme as I was struggling with rebuilding on Windows with updated dependencies (although github actions still builds cleanly).

This code also shows a nearly minimal example of adding a Command to agave.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,9 @@ pip install aqtinstall
 aqt install --outputdir C:\Qt 5.15.2 windows desktop win64_msvc2019_64
 ```
 
-Use vcpkg to install boost, tiff, glm. Make sure the vcpkg target triplet is x64-windows.
-
+Use vcpkg to install the following. Make sure the vcpkg target triplet is x64-windows.
 ```
-vcpkg install boost tiff glm --triplet x64-windows
+vcpkg install boost-log boost-filesystem glm zlib libjpeg-turbo liblzma --triplet x64-windows
 ```
 
 ```
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ pip install aqtinstall
 aqt install --outputdir C:\Qt 5.15.2 windows desktop win64_msvc2019_64
 ```
 
-Use vcpkg to install the following. Make sure the vcpkg target triplet is x64-windows.
+Use vcpkg (must use target triplet x64-windows) to install the following:
 ```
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -26,10 +26,9 @@ pip install aqtinstall
 aqt install --outputdir C:\Qt 5.15.2 windows desktop win64_msvc2019_64
 ```
 
-Use vcpkg to install boost, tiff, glm. Make sure the vcpkg target triplet is x64-windows.
-
+Use vcpkg to install the following. Make sure the vcpkg target triplet is x64-windows.
```
kmitcham, line 29: nit: 
"Make sure the vcpkg target triplet is x64-windows, and install the following:"

docs/conf.py
```diff
@@ -28,9 +28,11 @@
 # -- General configuration ---------------------------------------------------
 
 # Add any Sphinx extension module names here, as strings. They can be
-# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# extensions coming with Sphinx (named "sphinx.ext.*") or your custom
 # ones.
-extensions = ["m2r2"]
+extensions = [
```
kmitcham, line 33: Stylistically, I prefer the previous version, but it doesn't even rise to the level of a nit.  Was this an automated format?

renderlib/Logging.h
```diff
@@ -1,5 +1,8 @@
 #pragma once
 
+// remove after boost fixes issue in boost 1.78
```
kmitcham, line 3: Should this be a TODO, or maybe have a date added?  Or even a url to the issue tracker?

kmitcham (dismissed):

docs/conf.py
toloudis: Yes its an autoformat.  I'd been ignoring it for too long and am finally committing it.

griffinfujioka (dismissed):

Issue #15, opened 2021-06-10
Issue Building on Linux


Npikielny:
## Description
I tried building agave on linux using the steps outlined in the README, but I got the following error:

CMake Error at CMakeLists.txt:72 (find_package):
  By not providing "FindQt5.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Qt5", but
  CMake did not find one.

  Could not find a package configuration file provided by "Qt5" with any of
  the following names:

    Qt5Config.cmake
    qt5-config.cmake

  Add the installation prefix of "Qt5" to CMAKE_PREFIX_PATH or set "Qt5_DIR"
  to a directory containing one of the above files.  If "Qt5" provides a
  separate development package or SDK, be sure it has been installed.

## Environment

I am running Ubuntu 20.04.2.

toloudis:
This is a little bit particular to how (and where) you have installed Qt, but generally, the error message is giving the correct hints.

Setting a Qt5_DIR environment variable (to the directory of your Qt install, something like `~/Qt/5.15.2/gcc_64` ) before running cmake should help.

If you try this, let me know how it goes!

toloudis:
I'll update the README instructions to make this more clear.

toloudis:
This is because those missing libraries are git submodules.  They have to be pulled with one separate step that looks like this:
`git submodule update --init`
That should do the trick, and I'll make sure that goes into the readme also.

Npikielny:
Building worked! I ran the app from agave/build/agave (which I'm not sure is correct), and the gui works, but the shaders failed to compile:
GL_RENDERER: AMD Radeon RX 5700 XT (NAVI10, DRM 3.38.0, 5.8.0-55-generic, LLVM 11.0.0)
QOpenGLShader::compile(Fragment): 0:294(22): error: sampler arrays indexed with non-constant expressions are forbidden in GLSL 1.30 and later

toloudis:
Interesting -- I've not seen this error with agave before.   This may have something to do with the OpenGL version that the application requests when initializing. I can probably recommend a code change for you to try on your local copy of the code.

Npikielny:
Sounds great! What do I need to change?

toloudis:
A bit experimental but at this line:
https://github.com/allen-cell-animated/agave/blob/master/renderlib/renderlib.cpp#L26
change the major version to 4 and the minor version to 0 , rebuild and see if the shaders compile again.

[...]

PR #39, opened 2022-04-08
Fix/linux amd shader compilation error

### LINKED ISSUE #38: Path tracing broken on Linux
## Description

The path tracing fragment shader does not compile on Linux.
There are multiple locations where sampler arrays are indexed using integer variables.
Compilation fails with this error:
```
0000008 | 2022-04-07, 13:46:57.500381 [warning] - GLShader::compile(Fragment): 0:295(22): error: sampler arrays indexed with non-constant expressions are forbidden in GLSL 1.30 and later
```

Here's the complete [logfile.log](https://github.com/allen-cell-animated/agave/files/8445489/logfile.log).

## Reproduction

Start `agave`.

## Environment

OS: Ubuntu 20.04
OpenGL renderer string: AMD RENOIR (LLVM 13.0.0, DRM 3.41, 5.13.0-10039-tuxedo)
OpenGL core profile version string: 4.6 (Core Profile) Mesa 22.0.1 - kisak-mesa PPA

toloudis:
Fix errors with compiling shaders at runtime on linux with AMD drivers.
The bug fix is changing the shader version directives from v330 to v400 in the shader source code.

Also because of the nature of this fix, I am removing all occurrences of "v33" in file names and class names.  Lots of files get moved.  A few unused source files are deleted.

Fixes #38 and an issue mentioned in #15


renderlib/glsl/GLBasicVolumeShader.cpp
```diff
@@ -17,7 +17,7 @@ GLBasicVolumeShader::GLBasicVolumeShader()
 {
   vshader = new GLShader(GL_VERTEX_SHADER);
   vshader->compileSourceCode(R"(
-#version 330 core
+#version 400 core
```
toloudis, line 20: the changes in this directory of this type constitute the actual bug fix.  All other changes are simple renaming/moving files.

griffinfujioka (approved):

kmitcham (approved):

PR #63, opened 2022-06-08
add numbers to ROI sliders

toloudis:
The sliders that control the region of interest clipping now have value labels that tell you exactly at which pixel or slice the handle is positioned.

Before:
AICS GPU Volume Explorer 1 3 0 2022-06-08 10-25-17
After:
AICS GPU Volume Explorer 1 3 0 2022-06-08 10-24-37


toloudis:
> @toloudis only feedback is to give just a few pixels of space above the numbers so they don't touch the controls.

Will do.

griffinfujioka (approved):

kmitcham (approved):

PR #64, opened 2022-06-08
Feature/remove boost etc

toloudis:
This is a big PR covering a few key changes.  Hopefully the changes are not too interleaved and it's understandable.
1.  Remove `boost` as a dependency once and for all.  This should speed up build times, especially on Windows. The last boost components being used were boost-log (replaced with a library called `spdlog` - see `Logging.cpp`) and boost-filesystem (replaced with C++17's `std::filesystem`)
2. Small rendering changes in OpenGL to control the background alpha value better.  This is paving the way to enable users to output images with background transparency for easier compositing on slides etc.
3. Parallelize a couple of cpu-heavy tasks by way of the addition of a `parallel_for` loop that automatically creates threads, and blocks until all threads complete.  See `threading.cpp` for the parallel_for implementation. 
4. A change to the way pixel sampling is done in the core volume renderer, which should make better choices of the directions that the path tracer shoots its main camera rays. (all in `GLPTVolumeShader.cpp`)

### DIFF
--- a/agave_app/main.cpp
+++ b/agave_app/main.cpp
@@ -22,7 +22,8 @@ struct ServerParams
   // defaults
   ServerParams()
     : _port(1235)
-  {}
+  {
+  }
 };
 
 ServerParams
@@ -159,9 +160,14 @@ main(int argc, char* argv[])
 
     result = a.exec();
   } else {
-    agaveGui* w = new agaveGui();
-    w->show();
-    result = a.exec();
+    try {
+      agaveGui* w = new agaveGui();
+      w->show();
+      result = a.exec();
+
+    } catch (...) {
+      LOG_ERROR << "Exception caught in main";
+    }
   }
 
   renderlib::cleanup();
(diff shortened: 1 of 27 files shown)


griffinfujioka (dismissed):

agave_app/main.cpp
```diff
@@ -159,9 +160,14 @@ main(int argc, char* argv[])
...
     result = a.exec();
   } else {
-    agaveGui* w = new agaveGui();
-    w->show();
-    result = a.exec();
+    try {
+      agaveGui* w = new agaveGui();
+      w->show();
+      result = a.exec();
+
```
kmitcham, line 167: (nit) intentional whitespace?    Surprised a linter didn't catch this.

agave_app/main.cpp
```diff
@@ -159,9 +160,14 @@ main(int argc, char* argv[])
...
+    } catch (...) {
+      LOG_ERROR << "Exception caught in main";
```
kmitcham, line 169: Just continue on the error in main?  I suppose the program likely is crashing now anyway.

kmitcham (dismissed):

agave_app/main.cpp
toloudis: The curly brackets of the try block caused one more level of indentation. Is that what you're seeing?

PR #72, opened 2023-01-02
Render Window

toloudis:
Users would like more control over the final image output.  This PR provides a render window as a dialog box. 

This is the first revision and will likely undergo several modifications before release.

In the render window users can choose output resolution, time series, number of samples or for how long to render, and filename for output file.  Default output format is PNG (lossless rgba format).

AGAVE Render 2023-01-02 14-55-30

Behavior: 
Click "Render" to open the dialog.   Current view window will be echoed into the Render window, along with current resolution.
Select desired output resolution, number of samples, and click Render.   
Progress bar will indicate progress. A second progress bar is available for time series rendering.  Note that in between times of a time series, there will be a delay while the next data is loaded.
Autosave is enabled - frames will be saved to the indicated directory using the default filename with a numeric index following, like "frame_0000".  For time series this index will be incremented.
All settings from the Render dialog are saved to the JSON save format when you do "Save JSON".

Implementation detail:
The renderer runs on a thread and there is a complex interaction between the rendering thread and the main thread that controls the user interface of the dialog.  If testing, care should be taken to see if there are crashes during actions like pause or cancel or closing the dialog - anything that requires synchronization between the render thread and main thread.  My testing has shown it is stable but running on faster or slower computers may reveal issues.

Notes on the code changes:
* The GLView3d had some modifications to guarantee that OpenGL rendering is completely stopped while the Render Dialog is open, so that the GPU resources can be reused without having to destroy and reload memory.
* communication with render thread is done with Commands.  A big chunk of extra code around commands was introduced in order to be able to do binary encoding/decoding in the C++ code, using CommandBufferWriter.  This is a new in-process implementation of the same Command objects that are also used for binary transmission in server mode.
* attempt has been made to reuse the same Render Thread that is used in Server mode. However, the OpenGL context handling is somewhat complicated for all the different usage modes when trying to support cross-platform including headless EGL. Linux is currently broken and Server mode needs to be re-tested.


agave_app/GLView3D.cpp
```diff
@@ -366,3 +399,24 @@ GLView3D::captureQimage()
 
   return img;
 }
+
+void
+GLView3D::pauseRenderLoop()
+{
+  std::shared_ptr<CStatus> s = getStatus();
+// the CStatus updtes can cause Qt GUI work to happen,
```
toloudis, line 407: typo

agave_app/ViewerState.h
```diff
@@ -47,6 +47,20 @@ struct LightViewerState
   float m_width = 1.0f, m_height = 1.0f, m_distance = 10.0f;
 };
 
+struct CaptureState
+{
+  QString mOutputDir;
+  QString mFilenamePrefix;
```
toloudis, line 53: use std::string.  reserve QStrings for gui only

agave_app/agaveGui.cpp
```diff
@@ -121,9 +122,13 @@ agaveGui::createActions()
   m_toggleCameraProjectionAction->setStatusTip(tr("Toggle perspective and orthographic camera projection modes"));
   connect(m_toggleCameraProjectionAction, SIGNAL(triggered()), this, SLOT(view_toggleProjection()));
 
-  m_saveImageAction = new QAction(tr("&Save image..."), this);
+  m_saveImageAction = new QAction(tr("&Quick render..."), this);
   m_saveImageAction->setStatusTip(tr("Save the current render to an image file"));
   connect(m_saveImageAction, SIGNAL(triggered()), this, SLOT(saveImage()));
+
+  m_quickRenderAction = new QAction(tr("&Render..."), this);
```
toloudis, line 129: rename this, it's no longer a quick render

agave_app/agaveGui.cpp
```diff
@@ -309,6 +316,65 @@ agaveGui::saveImage()
   }
 }
 
+void
+agaveGui::onQuickRender()
```
toloudis, line 320: rename - not quick

agave_app/agaveGui.cpp
```diff
@@ -309,6 +316,65 @@ agaveGui::saveImage()
...
+  // copy of camera
+  // const appscene ref?
+  // const rendersettings ref?
+  // hand over Camera, AppScene and RenderSettings to RenderDialog?
+  // or a ViewerState?
+  // ViewerState st = appToViewerState();
+  // tell the renderdialog what the viewerstate currently is.
+  // renderdialog can turn it into a list of render commands
+  // share already loaded volume data with the RenderDialog's rendering resources
+  // (GL context sharing?)
```
toloudis, line 331: clean up the above comment

agave_app/main.cpp
```diff
@@ -145,10 +146,13 @@ main(int argc, char* argv[])
 
     // set to true to show windows, or false to run as a console application
     static const bool gui = false;
+    static const bool serverTest = true;
     if (gui) {
       MainWindow* _ = new MainWindow(server);
       _->resize(512, 512);
       _->show();
+    } else if (serverTest) {
+      StreamTestClient* client = new StreamTestClient(QUrl("ws://localhost:1235"), true);
```
toloudis, line 155: still need this?

agave_app/renderDialog.cpp
```diff
@@ -0,0 +1,843 @@
...
+  // the QImage is sent here from the thread.
+  // this is an incremental update of a render and our chance to update the GUI and state of our processing
+
+  if (req == nullptr) {
+    LOG_DEBUG << "render callback after no request processed";
+    return;
+  }
+  if (!m_renderThread || m_renderThread->isFinished() || m_renderThread->isInterruptionRequested()) {
+    LOG_DEBUG << "received sample after render terminated";
+  }
```
toloudis, line 429: why not return?

agave_app/renderer.cpp
```diff
@@ -165,76 +273,92 @@ Renderer::processRequest()
...
+    img = this->render();
 
     lastReq->setActualDuration(timer.nsecsElapsed());
 
     // in stream mode:
-    // if queue is empty, then keep firing redraws back to client.
-    // test about 100 frames as a convergence limit.
-    if (m_streamMode != 0 && m_myVolumeData.m_renderSettings->GetNoIterations() < 500) {
+    // if queue is empty, then keep firing redraws back to client, to build up iterations.
+    static const int MAX_ITERATIONS = 1024;
```
toloudis, line 295: is this putting a hard limit on samples for render?

kmitcham (approved):

agave_app/agaveGui.cpp
```diff
@@ -121,9 +122,13 @@ agaveGui::createActions()
   m_toggleCameraProjectionAction->setStatusTip(tr("Toggle perspective and orthographic camera projection modes"));
   connect(m_toggleCameraProjectionAction, SIGNAL(triggered()), this, SLOT(view_toggleProjection()));
 
-  m_saveImageAction = new QAction(tr("&Save image..."), this);
+  m_saveImageAction = new QAction(tr("&Quick render..."), this);
```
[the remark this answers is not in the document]
toloudis: I agree, it will probably be renamed again after some UX review.  It could also be named "Capture Viewport" or "Save current view" to distinguish it from "launch render dialog where I am going to take time to set up further image generation parameters"

agave_app/renderer.cpp
```diff
@@ -165,76 +273,92 @@ Renderer::processRequest()
...
+    // get the renderer we need
+    RenderGLPT* r = dynamic_cast<RenderGLPT*>(m_myVolumeData.m_renderer);
+    if (!r) {
+      LOG_ERROR << "Unsupported renderer: Renderer is not of type RenderGLPT";
+    }
+
+    m_ec.m_renderSettings = &r->getRenderSettings(); // m_myVolumeData.m_renderSettings;
     m_ec.m_renderer = this;
-    m_ec.m_appScene = m_myVolumeData.m_scene;
+    m_ec.m_appScene = r->scene(); // m_myVolumeData.m_scene;
```
frasercl, line 361: Will these commented-out assignments be useful for understanding or testing later?

toloudis: no, just cruft -- I'll remove.

frasercl (approved):
Been a long time with me and C++, but this looks good to go to me.

PR #73, opened 2023-01-13
Feature/zarr

toloudis:
Allow loading data from cloud if it is in OME-zarr format. 

This is a somewhat unpolished user experience but implements some of the most important functionality.

I add a "open from url" command.  Once a url (http(s)) is provided, then Agave will attempt to load ome-zarr metadata.  If such metadata is found, a dialog will pop up showing the multiresolution levels discovered.  Users will be provided a memory estimate of the selected level.  Users may also select a sub-region of interest in x,y,z.
Load Settings 2023-01-13 16-09-26

Once the selection is accepted, data will be downloaded and loaded into Agave.

Implementation notes:
* a new LoadSpec structure is introduced and it now replaces anywhere that an old file path and scene index was specified before. 
* new data structure and functions are introduced to load multiresolution data
* the LoadDialog is responsible for completely populating a LoadSpec which can then be passed to the FileReaders that do the data loading
* the old FileReaders are modified to use the new LoadSpec but do not handle multiresolution yet
* tensorstore is a google library that provides the underlying zarr loading support.  it's pretty heavy in that it adds a lot to the build time of agave
* a VERY small number of pixel formats and dimension configurations are currently supported for zarr.  Currently it's been tested with data from AIND and AICS that has TCZYX dimensions.


CMakeLists.txt
```diff
@@ -1,13 +1,14 @@
 cmake_minimum_required(VERSION 3.16)
```
toloudis, line 1: this file got autoformatted rather drastically. the modifications in here are mostly to simplify things, removing some extra flags

renderlib/FileReader.cpp
```diff
@@ -52,51 +56,79 @@ FileReader::loadFileDimensions(const std::string& filepath, uint32_t scene)
 {
   std::string extstr = getExtension(filepath);
 
-  if (extstr == ".tif" || extstr == ".tiff") {
+  if (filepath.find("http") == 0) {
```
kmitcham, line 59: These if cascades worry me, but are probably good enough for an initial version.

renderlib/command.cpp
```diff
@@ -535,57 +542,59 @@ SetTimeCommand::execute(ExecutionContext* c)
...
-    for (uint32_t i = 0; i < image->sizeC(); ++i) {
-      GradientData& lutInfo = c->m_appScene->m_material.m_gradientData[i];
-      lutInfo.convert(c->m_appScene->m_volume->channel(i)->m_histogram, image->channel(i)->m_histogram);
+  // we expect the scene volume dimensions to be the same; we want to preserve all view settings here.
+  // BUT we want to convert the old lookup tables to new lookup tables
+  // if we are preserving absolute transfer function settings
 
-      image->channel(i)->generateFromGradientData(lutInfo);
-    }
+  // assume sizeC is same for both previous image and new image!
```
kmitcham, line 568: if we are checking, are we still assuming?  Or am I misunderstanding something here?

renderlib/threading.cpp
```diff
@@ -50,3 +50,109 @@ parallel_for(size_t nb_elements, std::function<void(size_t start, size_t end)> f
...
+    }
+    // if the task is invalid, it means we are asked to abort:
+    if (!f.valid())
+      return;
+    // otherwise, run the task:
+    f();
+  }
+}
+
+#if 0
```
kmitcham, line 142: ?

kmitcham (dismissed):

renderlib/FileReader.cpp
toloudis: What is the concern here? Is it the repetition?
There's a future refactor where maybe the file type is determined much earlier and in one place only (in maybe a factory method) and then a derived-class reader implementation is selected that way.

renderlib/command.cpp
toloudis: Maybe I'm misusing the word assume, and I'm making the assumption explicit in code - "it's an error if they are different" but I still want to log the error in this app's particular logging system so that if it ever comes up I'll know what happened.  (But this isn't new code, it's just in the diff because the if statement containing this was removed.)

renderlib/FileReader.cpp
kmitcham: Both the repetition and the cascading if/else set that is hardcoded.  I'd prefer a lookup table somewhere independent of the rest of the stuff.  But maybe that is over-complicating things; the if/else cascade is very clear, at least.

renderlib/command.cpp
kmitcham: Maybe 'require sizeC to be the same for both previous image and new image'?

renderlib/FileReaderZarr.cpp
```diff
@@ -0,0 +1,340 @@
...
+    for (size_t b = 0; b < numPixels; ++b) {
+      *dataptr16 = (uint16_t)((src32[b] - lowest) / (highest - lowest) * 65535.0);
+      dataptr16++;
+    }
+    return 1;
+  } else {
+    LOG_ERROR << "Unexpected tiff pixel size " << srcBitsPerPixel << " bits";
+    return 0;
+  }
+  return 0;
```
frasercl, line 108: Unreachable `return` - could either remove or break the code out of the `else` block above

agave_app/agaveGui.cpp
```diff
@@ -382,94 +400,142 @@ agaveGui::saveJson()
...
+bool
+agaveGui::open(const std::string& file, const ViewerState* vs)
+{
+  LoadSpec loadSpec;
+  VolumeDimensions dims;
+
+  int sceneToLoad = vs ? vs->m_currentScene : 0;
+  int timeToLoad = vs ? vs->m_currentTime : 0;
+
+  if (file.find_first_of("http") == 0) {
```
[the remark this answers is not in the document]
toloudis: Oops!  THANK YOU. I have a startswith implementation somewhere in here (in one of the file readers) but was resisting pulling it out to a string utils file.  startswith is being introduced as standard in C++20 I believe, but this codebase is on C++17.

frasercl (approved):
My remaining comments are all minor cosmetic questions. This looks good to go functionally.

PR #80, opened 2023-02-02
Feature/local zarr

toloudis:
Some quick changes to allow loading zarr from local directories.
If the provided path either begins with http or contains "zarr" anywhere, it will be loaded as if it is ome-zarr.

The key implementation detail for these changes is providing the right type of KvStore Driver to tensorstore.

The front end needs to be cleaned up in a future update to how loading is initiated in the UI.

### DIFF
--- a/agave_app/agaveGui.cpp
+++ b/agave_app/agaveGui.cpp
@@ -470,7 +470,7 @@ agaveGui::open(const std::string& file, const ViewerState* vs)
   int sceneToLoad = vs ? vs->m_currentScene : 0;
   int timeToLoad = vs ? vs->m_currentTime : 0;
 
-  if (file.find("http") == 0) {
+  if (file.find("http") == 0 || file.find("zarr") != std::string::npos) {
     // read some metadata from the cloud and present the next dialog
     // if successful
 
--- a/agave_app/agaveGui.cpp
+++ b/agave_app/agaveGui.cpp
@@ -262,7 +262,7 @@ agaveGui::openUrl()
 {
   std::string urlToLoad = "";
   bool ok = false;
-  QString text = QInputDialog::getText(this, tr("Enter url or directory"), tr("Location"), QLineEdit::Normal, "", &ok);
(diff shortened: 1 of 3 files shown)


renderlib/FileReaderZarr.cpp
```diff
@@ -7,13 +7,42 @@
...
+}
+
+static nlohmann::json
+getKvStoreDriverParams(const std::string& filepath, const std::string& subpath)
+{
+  if (isCloud(filepath)) {
+    return { { "driver", "http" }, { "base_url", filepath }, { "path", subpath } };
+  } else {
+    // if file path does not end with slash then add one
+    // TODO maybe use std::filesystem::path for cross-platform?
```
toloudis, line 31: the slight ugliness that follows was done for expediency and it needs to be made more robust.  Expediency: so that it can be built and distributed internally for zarr testing.

BrianWhitneyAI (commented):
lgtm

agave_app/agaveGui.cpp
```diff
@@ -470,7 +469,7 @@ agaveGui::open(const std::string& file, const ViewerState* vs)
   int sceneToLoad = vs ? vs->m_currentScene : 0;
   int timeToLoad = vs ? vs->m_currentTime : 0;
 
-  if (file.find("http") == 0) {
+  if (file.find("http") == 0 || file.find("zarr") != std::string::npos) {
```
kmitcham, line 472: is the http case needed?  Could you just check or files ending in zarr?  Or do we have S3 objects that are zarrs named without the extension?

kmitcham (approved):

agave_app/agaveGui.cpp
toloudis: It's certainly true that the only thing agave can load over http is zarr, so if it's http we have to try the zarr loader. 
 
For local data, I'm just making a basic vague guess that if zarr is in the path, it should be one. Sometimes you can point to a subdirectory of a zarr store and still read a zarr.   I'm totally open to refining this later when I revisit data loading to make it all more streamlined from the front end.

Issue #82, opened 2023-02-07
implement new render UX


toloudis:
## Use Case

UX team has redesigned render dialog box. 

## Solution

Implement design from here: 
https://miro.com/app/board/uXjVP7GwOOM=/

## Alternatives

Literal translation of design is not required.  This will be a first iteration to get the re-layout done, new user flow, and find limits where the design may not be able to translate to Qt.

PR #85, opened 2023-02-20
Feature/render window UI

toloudis:
Revisions to render dialog to re-lay-out the controls and improve ux.
The basic input parameters to rendering have not changed - this is just cosmetic and user flow.  Some new feedback and ui dynamic updating is added.
Addresses #82 

single time:
AGAVE Render 2023-03-06 10-30-48
render finished:
AGAVE Render 2023-03-06 10-31-10
time series:
AGAVE Render 2023-03-06 10-34-24


toloudis:
The remaining main discrepancies with the design are a few things that are probably worth revisiting.
The preset resolution picker.  I still like it as a command button that you use to modify the numbers in the xy inputs.  It kinda felt like it was becoming more complicated to change it.
The button state at end of render.  It turns out that it’s hard to track “did the user change any values in any editable field” and so the current setting is a compromise that just lets you re-render when the render is done.  I was also feeling funny about having a new close button pop into existence when the user can already close the dialog by standard means.
Time increment seemed out of scope.  Why would a user skip time samples?  Either way this could easily be added later after this first batch of code changes.
Opening a file explorer also seemed out of scope as an extra feature that could be added later.

There’s also a user flow issue that came up:  
Every time you enter the dialog, I capture the current 3d view snapshot and stick that in the render image preview.  But the first time you enter the dialog, the XY resolution defaults to the window size but on subsequent times, it tries to remember the last XY render setting you had.  As a consequence, the render image preview gets out of sync with the XY resolution settings after the first time you render.  I considered fixing this by rescaling the snapshotted image and letterboxing it to fit the user-selected XY settings, but haven’t tried to implement that yet.

Windows seems to look better than Mac.  I’m kinda disappointed in that, but Qt is trying to keep the native Mac feel with the way it justifies and sizes the controls.  The font sizes are bugging me the most.

toloudis:
quick notes after meeting
remaining todos for implementation
- letterbox the preview window prior to rendering.  when changing aspect ratio, re-fit image to preview.
- add descriptive text about time series indices in filename
- make sure filename has no suffix when not time series
- position toolbar at lower right corner of image preview with inset (need to track window resizing)
- tooltips
- warn about overwrite as soon as start render is clicked

PR #92, opened 2023-05-26
Feature/docker updates post1.5

toloudis:
Revision to Docker image for v1.5 release with some dependency updates.
Key changes:
use nvidia-cuda docker base image with ubuntu 22.04, with opengl as addon installs.  Also force use gcc-10 for tensorstore compilation


Dockerfile
```diff
@@ -1,37 +1,47 @@
...
 RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null
-RUN apt-add-repository 'deb https://apt.kitware.com/ubuntu/ focal main'
+RUN apt-add-repository 'deb https://apt.kitware.com/ubuntu/ jammy main'
 RUN apt-get install kitware-archive-keyring
 RUN rm /etc/apt/trusted.gpg.d/kitware.gpg
 RUN apt-get update && apt-get install -y cmake
 
 # get python
-RUN apt-get install -y python3.9-dev python3-pip
+RUN apt-get install -y python3-pip
```
kmitcham, line 44: So the default python is fine now?

kmitcham (approved):

Dockerfile
toloudis: the base image is now ubuntu 22 which has python 3.10 as default

griffinfujioka (approved):

PR #107, opened 2023-06-01
Feature/raymarch server mode

toloudis:
**Pull request recommendations:**

- [ ] Name your pull request _your-development-type/short-description_. Ex: _feature/read-tiff-files_
- [ ] Link to any relevant issue in the PR description. Ex: _Resolves [gh-##], adds tiff file format support_
- [ ] Provide description and context of changes.
- [ ] Provide relevant tests for your feature or bug fix.
- [ ] Provide or update documentation for any feature added by your pull request.

Thanks for contributing!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,6 +196,10 @@ if(WIN32)
     DESTINATION ${INSTALLDIR}/imageformats/
   )
   install(FILES
+    $<TARGET_FILE:Qt6::QJpegPlugin>
+    DESTINATION ${INSTALLDIR}/imageformats/
+  )
+  install(FILES
     $<TARGET_FILE:Qt6::QWindowsVistaStylePlugin>
     DESTINATION ${INSTALLDIR}/styles/
   )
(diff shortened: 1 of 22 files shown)


CMakeLists.txt
toloudis, line 1: changes in this file because an extra Qt module had been left out of the distribution (a Qt plugin that can encode/decode jpeg)

agave_pyclient/agave_pyclient/agave.py
```diff
@@ -155,6 +155,13 @@ class AgaveRenderer:
     AgaveRenderer communicates with AGAVE running in server mode to perform GPU volume
     rendering.
 
+    Parameters
+    ----------
+    url: str
+        Full url to websocket server including port
+    mode: str
+        "pathtrace" or "raymarch" (pathtrace is default)
```
kmitcham, line 163: am I not seeing the code that gracefully fails for modes like 'path trace'?

toloudis: if you pass in an unrecognized mode name it will use "pathtrace".  That's handled on the other end where it decodes the string, renderlib/renderlib.cpp line 556.  At the python layer the string is just propagated through the pipe as-is.  I'll add some Python Typing stuff to decorate the function argument here.

PR #109, opened 2023-06-06
Compiler support/gcc 13

thewtex:
GCC 13 requires explicit `#include <cstdint>` for `uint32_t`, etc.


toloudis (approved):
LGTM!  Thanks for the contribution.

PR #110, opened 2023-06-06
Feature/reorg renderlib

toloudis:
This code change should have no effect on any functionality at all and is just moving some files.

Move io-related files into an `io/` subdirectory.
Move OpenGL graphics related files into a `graphics/` subdirectory.
This allows a bit of logical organization to help pave the way for future changes that may be graphics-only (or data-loading only). 
It also encapsulates some dependencies that are only relevant for io : tensorstore and libCZI.

For some reason, git recognized that some files were moved and others seem to show full diffs even though they are unchanged and only moved.


kmitcham (approved):
"I owe, I owe, its off to work I go..."

PR #112, opened 2023-06-08
Feature/js interface

toloudis:
Rewrite the agave web client with an eye toward integrating it in other viewers.

The previous code was not very reusable.  Note that the diff is so huge that you might as well ignore the old code.
There are a few remnants of the old implementation still here, which could be refined but will save that for a separate PR. 

This code change splits up the webclient directory into two sections: one containing the reusable agave websocket client, and one containing an example viewer that uses the client.

How it works:
Much like the python interface, we connect to an already running agave server using websockets.  We then encode small commands and send them to the agave server.  The agave server returns a stream of images back via binary websocket messages. Currently the images are encoded as jpg files.  This is slightly different than the python client, which receives images one at a time rather than streamed.  

The python use case is more designed for automation and generating a whole image from a batch of commands.  The web client is more designed for streaming interactive viewing.

A test viewer is implemented here, which simply deposits the received jpg data as a `img.src` which results in it being displayed on the web page.  The vast majority of the test viewer code is setting up a simple gui so we can test many of the agave commands.


agave_pyclient/agave_pyclient/agave.py
```diff
@@ -173,6 +173,8 @@ class AgaveRenderer:
     def __init__(self, url="ws://localhost:1235/", mode="pathtrace") -> None:
         self.cb = CommandBuffer()
         self.session_name = ""
+        if mode != "pathtrace" and mode != "raymarch":
+            mode = "pathtrace"
```
toloudis, line 177: bonus change: the python client will fix up the mode string if user supplied an invalid one.

webclient/public/index.ts
```diff
@@ -0,0 +1,941 @@
...
+    this.gui = new dat.GUI();
+    this._stream_mode = false;
+    this._stream_mode_suspended = false;
+    this.effectController = effectController;
+    this.enqueued_image_data = null;
+    this.waiting_for_image = false;
+    const wsUri = "ws://localhost:1235";
+    //const wsUri = "ws://dev-aics-dtp-001.corp.alleninstitute.org:1235";
+    //const wsUri = "ws://ec2-54-245-184-76.us-west-2.compute.amazonaws.com:1235";
+    this.agave = new AgaveClient(
```
toloudis, line 95: this is how an app would initialize a connection to agave

PR #116, opened 2023-06-20
allow to load tiff files produced by SCIFIO

toloudis:
SCIFIO injects its own flavor of metadata into tiff files, which was not recognized by AGAVE.  This allows those files to open.  I also received a TIFF file that has the INT sample type, and allowing that sample type to pass through seemed to just work.

### DIFF
--- a/renderlib/io/FileReaderTIFF.cpp
+++ b/renderlib/io/FileReaderTIFF.cpp
@@ -65,6 +65,33 @@ split(const std::string& s, char delim, std::vector<std::string>& elems)
   }
 }
 
+// multi lines split by newline
+// each line split by =
+std::map<std::string, std::string>
+splitToNameValuePairs(const std::string& s)
+{
+  std::vector<std::string> sl;
+  split(s, '\n', sl);
+
+  // split each string into name/value pairs,
+  // then look up as a map.
+  std::map<std::string, std::string> pairs;
+  for (int i = 0; i < sl.size(); ++i) {
+    std::vector<std::string> namevalue;
+    split(sl[i], '=', namevalue);
+    if (namevalue.size() == 2) {
+      pairs[namevalue[0]] = namevalue[1];
+    } else if (namevalue.size() == 1) {
+      pairs[namevalue[0]] = "";
+    } else {
+      // on error return empty map.
+      LOG_ERROR << "Unexpected name/value pair: " << sl[i];
+      return std::map<std::string, std::string>();
+    }
+  }
+  return pairs;
+}
+
 class ScopedTiffReader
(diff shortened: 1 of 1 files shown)


renderlib/io/FileReaderTIFF.cpp
```diff
@@ -65,6 +65,33 @@ split(const std::string& s, char delim, std::vector<std::string>& elems)
...
+{
+  std::vector<std::string> sl;
+  split(s, '\n', sl);
+
+  // split each string into name/value pairs,
+  // then look up as a map.
+  std::map<std::string, std::string> pairs;
+  for (int i = 0; i < sl.size(); ++i) {
+    std::vector<std::string> namevalue;
+    split(sl[i], '=', namevalue);
```
kmitcham, line 81: Is the data well-posed, or do you need to worry about whitespace?

renderlib/io/FileReaderTIFF.cpp
```diff
@@ -281,6 +296,103 @@ readTiffDimensions(TIFF* tiff, const std::string filepath, VolumeDimensions& dim
...
+      if (sizeC != nch) {
+        LOG_WARNING << "Inconsistent number of channels " << nch << " in SCIFIO TIFF: '" << filepath << "'";
+      }
+    } else {
+      LOG_WARNING << "Failed to read number of channels of SCIFIO TIFF: '" << filepath << "'";
+    }
+
+    for (uint32_t i = 0; i < sizeC; ++i) {
+      channelNames.push_back(std::to_string(i));
+    }
```
kmitcham, line 394: should you fail out here on fatal errors?  Or just expect to crash soon, having outputted diagnostic messages?

kmitcham (approved):

renderlib/io/FileReaderTIFF.cpp
toloudis: For now it's enough to have logged the error.  Most of the errors have some kind of graceful recovery that doesn't crash (for example: don't set the value we are trying to read, only load one channel instead of N channels, or return an empty image).    The docs for Agave say if something went wrong and you don't know why, look for the logfile first.

renderlib/io/FileReaderTIFF.cpp
toloudis: An extra whitespace trim wouldn't hurt here.  I believe this data is only ever written out programmatically so it's unlikely to have extra whitespace.  The problem is that I don't have a whole host of files to test with.  For most of these file format things, my philosophy is wait till someone tells me their file won't load and then decide whether to fix it.

PR #117, opened 2023-06-26
better typescript and naming for the webclient api

toloudis:
1. Rename api calls to be camelCase for js/ts rather than snake_case as a Python remnant.
2. Better typing across the board, removing the use of `any`, etc.
3. add eslint so that `npm run lint` works.


webclient/src/agave.ts
```diff
@@ -1,19 +1,27 @@
...
+  private onImage: (data: Blob) => void;
 
   constructor(
     url = "ws://localhost:1235/",
     rendermode = "pathtrace",
-    onOpen = () => {},
-    onJson = (json: any) => {},
-    onImage = (data: string) => {}
+    onOpen = () => {
+      0;
```
ShrimpCryptid, line 17: Similar note as on a different PR, might be cleaner to have this be a return statement

webclient/src/agave.ts
```diff
@@ -1,19 +1,27 @@
 import { CommandBuffer, COMMANDS } from "./commandbuffer";
 
+export type JSONValue = string | number | boolean | { [x: string]: JSONValue } | Array<JSONValue>;
+
 export class AgaveClient {
   private binarysocket0: WebSocket;
```
ShrimpCryptid, line 6: Should this variable use camelCase?

webclient/src/agave.ts
```diff
@@ -551,7 +556,7 @@ export class AgaveClient {
    * @param g The green value between 0 and 1
    * @param b The blue value between 0 and 1
    */
-  background_color(r: number, g: number, b: number) {
+  backgroundColor(r: number, g: number, b: number) {
```
ShrimpCryptid, line 559: Any reason for these setter methods being `backgroundColor` vs. `setBackgroundColor`?

webclient/src/commandbuffer.ts
```diff
@@ -103,20 +104,15 @@ export class CommandBuffer {
...
       const nArgsExpected = signature.length - 1;
       // for each arg:
       if (command.length - 1 !== nArgsExpected) {
-        console.error(
-          "BAD COMMAND: EXPECTED " +
-            nArgsExpected +
-            " args and got " +
-            (command.length - 1)
-        );
+        console.error("BAD COMMAND: EXPECTED " + nArgsExpected + " args and got " + (command.length - 1));
```
ShrimpCryptid, line 115: Super minor nitpick, could be cleaner as a template literal

ShrimpCryptid (approved):
Some minor formatting nitpicks, but this is a nice change especially after what we were talking about in the Retro.

webclient/src/agave.ts
toloudis: Agreed - I will fix these.

webclient/src/agave.ts
toloudis: I'm not going to change this right now because I'd have to change it in many many places.  The basic reasoning is that early on, almost every command was a "set" command so why bother saying set all the time.  But gradually I forgot and added some with the word set so it's an inconsistent mix now.   Good idea to clean it up in a future PR.  Probably will make some "set" command names that just alias to the old ones so as not to break anything.

webclient/src/agave.ts
toloudis: I'll just change it to `socket`

ShrimpCryptid (approved):
Looks good!

Issue #122, opened 2023-07-11
increase the max pixel resolution for render output


toloudis:
16384 x 16384

Future:
consider dpi x inches as a way to enter this

toloudis:
TODO : error handling for out of memory situations

PR #129, opened 2023-08-24
feat: Increased max render resolution

ShrimpCryptid:
Resolves https://github.com/allen-cell-animated/agave/issues/122!

Super tiny PR to increase the maximum resolution of images from 4096x4096 to 16384x16384.

I validated this by generating an image with a resolution of 4097x3704. (I also tried making a full-resolution image and AGAVE crashed and nearly took my computer with it.)

### DIFF
--- a/agave_app/renderDialog.cpp
+++ b/agave_app/renderDialog.cpp
@@ -2,10 +2,10 @@
 #include "renderer.h"
 
 #include "renderlib/AppScene.h"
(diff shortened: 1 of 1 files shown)


toloudis (approved):

toloudis:
A future thing is to catch that type of out-of-memory error and see if we can handle it more gracefully.  Or: write code to render very large images in smaller tiles that fit in memory.

kmitcham (approved):

Issue #102, opened 2023-05-30
loading dialog: use different titles depending on context


toloudis:
## Use Case

reassure the user they are performing the correct task.

## Solution

Loading Dialog: Change titles of all dialogs to reflect entry button clicked. This will help reassure the user they are performing the correct task. e.g. (alternate option (file type (e.g. .zarr, .tiff, etc.) Load Settings)
File Load Settings
Directory Load Settings
URL Load Settings

PR #130, opened 2023-08-25
feat: Rename loading dialog per context

ShrimpCryptid:
Resolves https://github.com/allen-cell-animated/agave/issues/102, "loading dialog: use different titles depending on context."

## Solution
- Uses an enum to determine which filetype is being loaded, and changes the title of the loading dialog based on the type.
- Added extra logic when loading recent files to determine the file type, since the information is not stored.

Possible names:
- "Load Settings" (for unknown types)
- "File Load Settings"
- "URL Load Settings"
- "Directory Load Settings"
- "JSON Load Settings" (this usually is not shown because JSON doesn't trigger the dialog).

## Screenshots 
image
image

### DIFF
--- a/agave_app/agaveGui.cpp
+++ b/agave_app/agaveGui.cpp
@@ -19,7 +19,6 @@
 #include "StatisticsDockWidget.h"
 #include "TimelineDockWidget.h"
 #include "ViewerState.h"
-#include "loadDialog.h"
 #include "renderDialog.h"
 
 #include <QAction>
@@ -261,7 +260,7 @@ agaveGui::open()
   QString file = QFileDialog::getOpenFileName(this, tr("Open Volume"), dir, QString(), 0, options);
 
   if (!file.isEmpty()) {
-    if (!open(file.toStdString(), "File Load Settings")) {
+    if (!open(file.toStdString(), LoadingDialogType::FILE)) {
       showOpenFailedMessageBox(file);
     }
   }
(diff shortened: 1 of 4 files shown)


agave_app/agaveGui.cpp
```diff
@@ -803,8 +802,19 @@ agaveGui::openRecentFile()
...
+      QString extension = path.sliced(path.lastIndexOf("."));
+      if (path.indexOf("http") == 0) {
+        type = LoadingDialogType::URL;
+      } else if (extension.indexOf(".zarr") == 0) {
+        type = LoadingDialogType::DIRECTORY;
+      } else if (extension.indexOf(".json") == 0) {
+        type = LoadingDialogType::JSON;
+      } else {
+        type = LoadingDialogType::FILE;
+      }
```
toloudis, line 815: nit: this could be a function getLoadDialogTypeFromFilePath

toloudis: ... could the load dialog figure this out for itself?

toloudis (approved):
one possible optional revision but overall LGTM

kmitcham (approved):

agave_app/agaveGui.cpp
ShrimpCryptid: You're right, it could. I'll move the code logic for it.

toloudis (approved):
LGTM!

toloudis (approved):
still LGTM

PR #131, opened 2023-08-26
Feature/manipulators

toloudis:
This PR introduces a code framework for in-viewport manipulators.
At a high level, here are the important strategic changes:
* move much of the GLView3D implementation down into renderlib, in a new class called ViewportWindow. We want as little app implementation as possible up in the agave_app Qt code.  This makes more and more of renderlib potentially reusable outside of Qt.
* introduce the Gesture framework (borrowed EXTENSIVELY from Max Liani's series of blog posts in 2021).  Gesture tracks mouse events (Gesture::Input) and also provides a simple immediate mode drawing framework for drawing the manipulators as colored lines or triangles (Gesture::Draw).
* introduce Manipulator classes to draw and do user interaction.
* for GLView3D to work well with Gesture, we needed to change the redraw loop slightly. Gesture is made for realtime constantly updating viewports.
* an audit of all the transforms in the app was done, to try to always be aware of coordinate systems and remove unnecessary coordinate flips.
* A "move" manipulator is implemented and tested but not completely integrated yet.  Also the very beginning of a rotate tool.

Agave's camera controls are now implemented within this framework, and as a demonstration of new capabilities, pressing A in viewport will animate the camera to center the view.
Apart from this, the rest of AGAVE should be expected to work as before.

Future work:
Next steps are to complete a rotate tool, add scroll-to-zoom, and consider UX to complete the integration of in-viewport interaction with respect to user flow. (when to hide/show the tools, when to initiate interactivity, and how to get tool updates to be shown in the rest of the UI).   Implementation details most likely will include having some kind of SceneObject class with a position and rotation that can be edited.


renderlib/gesture/gesture.cpp
```diff
@@ -0,0 +1,561 @@
...
+configure_3dDepthTested(SceneView& sceneView, Gesture::Graphics& graphics)
+{
+  auto& shader = graphics.shader;
+
+  glm::mat4 v(1.0);
+  sceneView.camera.getViewMatrix(v);
+  glm::mat4 p(1.0);
+  sceneView.camera.getProjMatrix(p);
+
+  // glm::mat4 s = glm::scale(glm::mat4(1.0), glm::vec3(1.0, -1.0, 1.0));
```
toloudis, line 96: delete commented line

toloudis:
As this is a big PR, here's my suggestion of how to get through it short of splitting it apart.

The big files are gesture.h/cpp, Manipulator.h/cpp, and ViewportWindow.h/cpp.   (CCamera.h/cpp also got some significant code added but it's not very complicated.)

The remaining changes are relatively small / easy to understand in comparison. 

One way to get through the big files is to read the comments first.  I think it's pretty well commented.  Also because it's derived from some blog posts, the blog goes into more depth explaining the motivation.

The big idea is "immediate mode" - on every redraw we are updating the input state (including picking), updating camera, evaluating manipulator actions, recalculating their drawing primitives, and then redrawing.

renderlib/gesture/gesture.h
```diff
@@ -0,0 +1,388 @@
...
+  struct Input
+  {
+    // User Input
+    static constexpr size_t kButtonsCount = 3;
+
+    // During app initialization query the OS accessibility settings how the user configured the
+    // double-click duration. Developer: never hardcode this time to something that feels
+    // right to you.  The initial value is just a nonzero default that is meant to be overriddent.
+    double doubleClickTime = 0.5; //< Mouse double-click time in seconds.
+    void setDoubleClickTime(double seconds) { doubleClickTime = seconds; }
```
ShrimpCryptid, line 36: The blog posts mentions querying the OS APIs to set this, did you ever end up implementing that?

toloudis: Yep, you can see that code in GLView3D.cpp

renderlib/gesture/gesture.h
```diff
@@ -0,0 +1,388 @@
...
+    // pointers to setup each draw pass.
+    enum class CommandSequence : int
+    {
+      k3dDepthTested = 0, //< for viewport depth compositing 3d GUI elements
+      k3dStacked = 1,     //< for 3d GUI elements overlay
+      k2dScreen = 2       //< for topmost 2d elements (widgets and buttons)
+    };
+    static constexpr int kNumCommandsLists = 3;
+
+    struct Command
```
ShrimpCryptid, line 158: Would be nice to have a comment here explaining what a Command is? I found it while looking through the blog but may not be immediately obvious for people touching this code for the first time.

renderlib/gesture/gesture.h
```diff
@@ -0,0 +1,388 @@
...
+        , g(c.y)
+        , b(c.z)
+        , a(opacity)
+        , s(selectionCode)
+      {
+      }
+
+      float x, y, z;    //< position
+      float u, v;       //< texture coordinates
+      float r, g, b, a; //< draw color (multiplies texture)
```
ShrimpCryptid, line 217: Nit: Should this go before the constructors?

renderlib/gesture/gesture.h
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    enum class LoopEntry : int
+    {
+      kContinue = 0,
+      kClose = 1
+    };
+    // Create a polyline (open or close) by extending a line created with
+    // addLine.
+    inline void extLine(const VertsCode& v, LoopEntry loop = LoopEntry::kContinue)
```
ShrimpCryptid, line 355: Nit: Maybe worth expanding this to be `extendLine`, since `ext` is a bit ambiguous

renderlib/gesture/gesture.cpp
```diff
@@ -0,0 +1,559 @@
...
+  glUniformMatrix4fv(shader->m_loc_proj, 1, GL_FALSE, glm::value_ptr(p * v));
+  check_gl("set proj matrix");
+
+  glEnable(GL_DEPTH_TEST);
+  check_gl("enable depth test");
+}
+
+// The second pass is still about 3d geometry, only this time I want it to be
+// drawn on top, without depth test. These two passes shares in common the same
+// projection matrix as the rest of the scene. 3d manipulators shown earlier are examples.
```
ShrimpCryptid, line 105: Appreciated these comments! Not sure if they were from the blog but it was helpful in understanding how the Pipeline worked.

renderlib/gesture/gesture.cpp
```diff
@@ -0,0 +1,559 @@
...
+uint32_t
+selectionRGB8ToCode(const uint8_t* rgba)
+{
+  // ignores 4th component (== 0)
+  uint32_t code = (uint32_t(rgba[0]) << 0) | (uint32_t(rgba[1]) << 8) | (uint32_t(rgba[2]) << 16);
+  return code == 0xffffff ? Gesture::Graphics::SelectionBuffer::k_noSelectionCode : code;
+}
+
+bool
+Gesture::Graphics::pick(SelectionBuffer& selection, const Gesture::Input& input, const SceneView::Viewport& viewport)
```
ShrimpCryptid, line 452: This was a fun section to read through! :D

renderlib/MathUtil.h
ShrimpCryptid, line 1: We talked about this over video call but some of the operations here should probably have their own unit tests!

renderlib/MathUtil.cpp
ShrimpCryptid, line 1: We talked about this over video call but some of the operations here should probably have their own unit tests!

renderlib/MathUtil.h
```diff
@@ -0,0 +1,110 @@
...
+  {
+  }
+  AffineSpace3f inverse() const;
+};
+
+inline glm::vec3
+xfmVector(const struct LinearSpace3f& xfm, glm::vec3 p)
+{
+  return xfm.vx * p.x + xfm.vy * p.y + xfm.vz * p.z;
+}
```
ShrimpCryptid, line 55: What does `xfm` mean here?

renderlib/MathUtil.h
```diff
@@ -0,0 +1,110 @@
...
+///
+/// This function solves for distance lineT.
+inline float
+linePlaneIsectT(const glm::vec3& planeP, const glm::vec3& planeN, const glm::vec3& lineP, const glm::vec3& lineV)
+{
+  float det = dot(lineV, planeN);
+  float lineT = (abs(det) < 1e-19f ? 0.0f : dot(planeP - lineP, planeN) / det);
+
+  return lineT;
+}
```
ShrimpCryptid, line 77: Is lineV expected to be normalized?

renderlib/MathUtil.h
toloudis: Yes, just shorthand for transform.  It's a word that gets used so much it becomes second nature to abbreviate it.

renderlib/MathUtil.h
toloudis: It doesn't need to be normalized because the returned value is a "t" value that is intended to be multiplied with lineV.  I think in this case the planeN is expected to be normalized.

renderlib/SceneView.h
```diff
@@ -0,0 +1,62 @@
...
+
+      void extend(const glm::ivec2& p)
+      {
+        lower = glm::min(lower, p);
+        upper = glm::max(upper, p);
+      }
+
+      glm::ivec2 size() const { return upper - lower; }
+
+      bool empty() const { return size().x < 0 || size().y < 0; }
```
ShrimpCryptid, line 44: Nit: isEmpty vs. empty?

ShrimpCryptid: Also should this be <= comparison?

renderlib/graphics/glsl/GLPTAccumShader.cpp
```diff
@@ -101,18 +101,4 @@ GLPTAccumShader::setShadingUniforms()
...
-  glm::vec3 eye(camera.m_From.x, camera.m_From.y, camera.m_From.z);
-  glm::vec3 center(camera.m_Target.x, camera.m_Target.y, camera.m_Target.z);
-  glm::vec3 up(camera.m_Up.x, camera.m_Up.y, camera.m_Up.z);
-  glm::mat4 cv = glm::lookAt(eye, center, up);
-  glm::mat4 cp = glm::perspectiveFov(vfov, w, h, camera.m_Near, camera.m_Far);
-
-  // glUniform3fv(uCameraPosition, 1, glm::value_ptr(camera.position));
-  // glUniformMatrix4fv(uProjectionMatrix, 1, GL_FALSE, glm::value_ptr(cp));
-  // glUniformMatrix4fv(uModelViewMatrix, 1, GL_FALSE, glm::value_ptr(cv * modelMatrix));
-  // glUniformMatrix4fv(uInverseModelViewMatrix, 1, GL_FALSE, glm::value_ptr(glm::inverse(cv * modelMatrix)));
```
ShrimpCryptid, line 117: Why was this code section removed? (this is not a leading question I am just unsure)

toloudis: all that math was actually doing nothing.  It was computing the transforms but never setting them anywhere.  Dead code.

renderlib/SceneView.h
toloudis: precedent for using the name `empty()` is in C++ STL containers.   But I get it, it's a controversial choice.  I guess empty is in the sense of an adjective and not a verb.

renderlib/Manipulator.h
```diff
@@ -0,0 +1,225 @@
...
+#include "AppScene.h"
+#include "MathUtil.h"
+#include "RenderSettings.h"
+#include "gesture/gesture.h"
+
+#include <assert.h>
+#include <inttypes.h>
+#include <vector>
+
+struct ManipulationTool
```
ShrimpCryptid, line 12: Question: When is it good practice to use classes vs. structs?

renderlib/Manipulator.cpp
```diff
@@ -0,0 +1,574 @@
...
+    glm::vec3 p1 = center + xaxis * cosf(theta1) + yaxis * sinf(theta1);
+
+    gesture.graphics.addLine(Gesture::Graphics::VertsCode(p0, color, opacity, code),
+                             Gesture::Graphics::VertsCode(p1, color, opacity, code));
+  }
+}
+
+// does not draw a flat base
+void
+drawCone(Gesture& gesture,
```
ShrimpCryptid, line 34: Maybe worth commenting that the `xaxis`, `yaxis`, `zaxis` variables also control height/radius of the cone?

renderlib/Manipulator.cpp
```diff
@@ -0,0 +1,574 @@
...
+         glm::vec3 base,
+         glm::vec3 xaxis,
+         glm::vec3 yaxis,
+         glm::vec3 zaxis,
```
ShrimpCryptid, line 38: Nit: camel case

renderlib/Manipulator.cpp
```diff
@@ -0,0 +1,574 @@
...
+        motion = lineLineNearestPoint(p, ref.vy, l, l1) - lineLineNearestPoint(p, ref.vy, l, l0);
+        break;
+
+      case MoveTool::kMoveZ:
+        motion = lineLineNearestPoint(p, ref.vz, l, l1) - lineLineNearestPoint(p, ref.vz, l, l0);
+        break;
+
+      case MoveTool::kMoveYZ:
+        // To move on two axis we compute the distance between two line-plane
+        // interections. The plane is that formed by the two axis.
```
ShrimpCryptid, line 179: Small typo in intersections

renderlib/Manipulator.h
toloudis: This is how I think about it: because structs have all members public by default, they are much simpler for small "plain old data" POD data types.  I would generally use structs for that and classes for anything else that is likely to have methods and need to possibly have public/private/protected stuff. If I am making a struct and need something to be private, I would pause and evaluate my choice.   In this case I can't really think of a good reason why a ManipulationTool should prefer struct over class. I actually think of it more as a class.  But in C++ the distinction is pretty minor.

renderlib/CCamera.h
```diff
@@ -545,7 +574,9 @@ class CCamera
     Update();
   }
 
-  float GetHorizontalFOV_radians()
+  float getHalfHorizontalAperture() const { return tan(this->GetHorizontalFOV_radians() * 0.5f); }
+
+  float GetHorizontalFOV_radians() const
```
ShrimpCryptid, line 579: capitalization?

renderlib/CCamera.h
```diff
@@ -573,19 +607,93 @@ class CCamera
...
-    // a code audit (and some unit tests) could clean this all up,
-    // starting in part by keeping these unflipped.
     if (m_Projection == PERSPECTIVE) {
-      projMatrix =
-        glm::perspectiveFov(vfov, w, h, m_Near, m_Far) * glm::scale(glm::mat4(1.0), glm::vec3(1.0, -1.0, 1.0));
+      projMatrix = glm::perspectiveFov(vfov, w, h, m_Near, m_Far);
     } else {
       projMatrix = glm::ortho(
-        -(w / h) * m_OrthoScale, (w / h) * m_OrthoScale, 1.0f * m_OrthoScale, -1.0f * m_OrthoScale, m_Near, m_Far);
+        -(w / h) * m_OrthoScale, (w / h) * m_OrthoScale, -1.0f * m_OrthoScale, 1.0f * m_OrthoScale, m_Near, m_Far);
```
ShrimpCryptid, line 614: Nice, glad to see these negations removed

renderlib/CCamera.h
```diff
@@ -573,19 +607,93 @@ class CCamera
...
+  float duration; //< animation total time
+  float time;     //< animation current time
+  CameraModifier mod;
+};
+
+struct Gesture;
+
+extern bool
+cameraManipulation(const glm::vec2 viewportSize,
+                   // const TimeSample& clock,
```
ShrimpCryptid, line 673: Remove?

renderlib/CCamera.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+  if ((button.modifier & Gesture::Input::kShift) == 0) {
+    // Exponential motion, the closer to the target, the slower the motion,
+    // the further away the faster.
+    glm::vec3 v_scaled = v * expf(-dragDist * dragScale);
+    motion = v_scaled - v;
+    targetMotion = glm::vec3(0);
+  } else {
+    // Linear motion. We move position and target at once. This mode allows the user not
+    // to get stuck with a camera that doesn't move because the postion got to close to
```
ShrimpCryptid, line 142: too vs. to, minor typo in position

renderlib/CCamera.cpp
```diff
@@ -0,0 +1,263 @@
...
+      cameraManipulationDolly(viewportSize, gesture.input.mbs[Gesture::Input::kButtonLeft], camera, cameraMod);
+  }
+  // Camera Tumble
+  // hasButtonAction with zero actions must be last to check for, since it will match anything
+  else if (gesture.input.hasButtonAction(Gesture::Input::kButtonLeft, 0 /*Gesture::Input::kAlt*/)) {
+    cameraEdit = true;
+    Gesture::Input::Button& button = gesture.input.mbs[Gesture::Input::kButtonLeft];
+
+    // LOG_DEBUG << "Pressed pos " << glm::to_string(button.pressedPosition) << ", cursor "
+    //           << glm::to_string(gesture.input.cursorPos) << ", drag " << glm::to_string(button.drag);
```
ShrimpCryptid, line 224: Remove?

renderlib/ViewerWindow.cpp
```diff
@@ -0,0 +1,222 @@
...
+    case 1:
+      LOG_DEBUG << "Set OpenGL pathtrace Renderer";
+      m_renderer.reset(new RenderGLPT(m_renderSettings));
+      m_renderSettings->m_DirtyFlags.SetFlag(TransferFunctionDirty);
+      break;
+    case 2:
+      LOG_DEBUG << "Set OpenGL pathtrace Renderer";
+      m_renderer.reset(new RenderGLPT(m_renderSettings));
+      m_renderSettings->m_DirtyFlags.SetFlag(TransferFunctionDirty);
+      break;
```
ShrimpCryptid, line 210: Are both these cases intentionally the same?

ShrimpCryptid (approved):
Um! So I at least _looked_ at all the code changes, but I'm not sure how much high-level feedback I'm able to give, and especially so given the scale of this PR. I'm approving it for now because I know these changes need to go through...

This definitely would have benefit from being split into several smaller PRs, because of the time commitment and the quality of discussion we're able to have about each piece. Getting through this PR took me a bit over full workday, and I'm still not sure I understand all the pieces of it. ((I regret what I said earlier about this size being fine :X))

One thing I noticed from playing around with the UI is that the way Gestures is set up doesn't seem to support changing what modifiers you're using in the middle of a click/drag operation. That might be a UX question to consider down the line.

toloudis:
> This definitely would have benefit from being split into several smaller PRs, because of the time commitment and the quality of discussion we're able to have about each piece. 

Understood.

> One thing I noticed from playing around with the UI is that the way Gestures is set up doesn't seem to support changing what modifiers you're using in the middle of a click/drag operation. That might be a UX question to consider down the line.

I think you could do this by looking at the gesture input state in the action and draw functions. You'd just have to implement the Manipulator to check for what info it cares about.

renderlib/ViewerWindow.cpp
toloudis: Yes intentional, I think there is some legacy reason for it where "2" used to be valid.

ShrimpCryptid: Would it be better to merge the two cases then?
```
case 1:
case 2:
  <code here>
```

PR #139, opened 2023-09-07
Feature/rotatetool

toloudis:
Adds a basic rotate manipulator.  
Also a bonus bug fix for clearing the pick buffer properly when it gets resized, so that the pick buffer doesn't return strange errors when there is nothing to pick.

Some refinements still to make due to only using "fake" object for rotations.  When this is more fully integrated there may be some extra "polish" features added to the manipulator.  This captures a subset of the functionality from Max Liani's Gesture rotation tool.

Next immediate work: connect this to the area light more correctly.  Center the rotate tool on the volume, and use it to rotate the area light's direction vector.

### DIFF
--- a/renderlib/RotateTool.cpp
+++ b/renderlib/RotateTool.cpp
@@ -36,19 +36,11 @@ getDraggedAngle(const glm::vec3& vN, const glm::vec3& p, const glm::vec3& l, con
   // if we can't intersect the planes properly, then we must be on-axis (plane perpendicular to view plane)
   // and we must calculate angle another way (TODO)
   if (p0 == p1 || axisIsOrthogonal) {
-    // find angle between (p,l0) and (p,l1)
-    // using as N, the view direction
-    // glm::vec3 v0 = normalize(p - l0);
-    // glm::vec3 v1 = normalize(p - l1);
-    // float angle = getSignedAngle(v0, v1, vN);
-    // return angle;
-
     // we want a linear measure of the amount of drag along the line of the ring
     glm::vec3 projectionAxis = cross(vN, globalAxis);
     glm::vec3 delta = l1 - l0;
     float projection = dot(delta, projectionAxis);
-    LOG_DEBUG << "on-axis rotation " << projection;
-    return projection;
+    return projection * glm::two_pi<float>();
   }
 
   // get angle between (p,p0) and (p,p1)
--- a/renderlib/RotateTool.cpp
+++ b/renderlib/RotateTool.cpp
@@ -8,6 +8,45 @@ struct ManipColors
   static constexpr glm::vec3 bright = { 1.0f, 1.0f, 1.0f };
 };
 
+static float
+getSignedAngle(const glm::vec3& v0, const glm::vec3& v1, const glm::vec3& vN)
+{
+  // get signed angle between the two vectors using (1,0,0) as plane normal
+  // atan2 ( (v0 X v1) dot (vN), v0 dot v1 )
+
+  float dp = dot(v0, v1);
+  float angle = atan2(dot(cross(v0, v1), vN), dp);
+  return angle;
+}
+
+static float
+getDraggedAngle(const glm::vec3& vN, const glm::vec3& p, const glm::vec3& l, const glm::vec3& l0, const glm::vec3& l1)
+{
+  // axis is represented by vN, and a point in plane p (the center of our circle)
+  // line l-l0 is the ray of the initial mouse click
+  // line l-l1 is the ray of the current mouse position
+  // project our drag pts into plane of circle, then find the angle between
+  // the two vectors
+  glm::vec3 p0 = linePlaneIsect(p, vN, l, l0);
+  glm::vec3 p1 = linePlaneIsect(p, vN, l, l1);
+  // if we can't intersect the planes properly, then we must be on-axis (plane perpendicular to view plane)
+  // and we must calculate angle another way (TODO)
+  if (p0 == p1) {
+    // find angle between (p,l0) and (p,l1)
+    // using as N, the view direction
+    glm::vec3 v0 = normalize(p - l0);
+    glm::vec3 v1 = normalize(p - l1);
+    float angle = getSignedAngle(v0, v1, vN);
+    return angle;
+  }
+
+  // get angle between (p,p0) and (p,p1)
+  glm::vec3 v0 = normalize(p - p0);
+  glm::vec3 v1 = normalize(p - p1);
+  float angle = getSignedAngle(v0, v1, vN);
+  return angle;
+}
+
 void
 RotateTool::action(SceneView& scene, Gesture& gesture)
 {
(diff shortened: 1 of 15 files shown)


renderlib/RotateTool.cpp
```diff
@@ -0,0 +1,367 @@
...
+static float
+getSignedAngle(const glm::vec3& v0, const glm::vec3& v1, const glm::vec3& vN)
+{
+  // get signed angle between the two vectors using (1,0,0) as plane normal
+  // atan2 ( (v0 X v1) dot (vN), v0 dot v1 )
+
+  float dp = dot(v0, v1);
+  float angle = atan2(dot(cross(v0, v1), vN), dp);
+  return angle;
+}
```
ShrimpCryptid, line 22: I was looking up formulas for getting the angle between two 3D vectors and it looks like a lot of them do:
```
arccos( (v0 dot v1) / (|v0| * |v1|)
```
Is this calculating the angle between the vectors when projected onto the plane given by vN?

renderlib/RotateTool.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+static float
+getDraggedAngle(const glm::vec3& vN, const glm::vec3& p, const glm::vec3& l, const glm::vec3& l0, const glm::vec3& l1)
+{
+  glm::vec3 globalAxis = normalize(p - l);
+  bool axisIsOrthogonal = abs(dot(globalAxis, vN)) < s_orthogonalThreshold;
+
+  // axis is represented by vN, and a point in plane p (the center of our circle)
+  // line l-l0 is the ray of the initial mouse click
+  // line l-l1 is the ray of the current mouse position
```
ShrimpCryptid, line 32: Is this convention for lines? Could these variables be named in a more descriptive way?

renderlib/RotateTool.cpp
```diff
@@ -0,0 +1,367 @@
...
+  // project our drag pts into plane of circle, then find the angle between
+  // the two vectors
+  glm::vec3 p0 = linePlaneIsect(p, vN, l, l0);
+  glm::vec3 p1 = linePlaneIsect(p, vN, l, l1);
+  // if we can't intersect the planes properly, then we must be on-axis (plane perpendicular to view plane)
+  // and we must calculate angle another way (TODO)
```
ShrimpCryptid, line 38: Still TODO?

renderlib/RotateTool.cpp
toloudis: The comment is wrong about (1,0,0); It should say `using vN as plane normal`.  I'll fix that.

The acos(dot product) is resulting in a function that doesn't know which direction the angle goes in.  acos(dot(x,y)) == acos(dot(y,x)).  We want a SIGNED angle based on sweeping a rotation FROM one vector to the other.   The cross product is order-dependent and cross(a,b) = - cross(b,a) which gives us the proper sense of rotating clockwise vs counterclockwise.   It does the projection just in case the vectors v0,v1 are not in plane.  vN also gives us the sense of which direction should be considered positive.    

cross product is proportional to sin, and dot product is proportional to cos, therefore we can do an atan with both sin and cos as parameters.

renderlib/RotateTool.cpp
toloudis: nope! will remove!

renderlib/RotateTool.cpp
toloudis: It's a convention in this code.  The way l, l0, and l1 are set up is: l is the "camera origin", l0 is the first point clicked, and l1 is the second point (where the mouse was currently dragged to).  I hope the previous comments where those vars are defined should describe them.  
I could rename the vars tho...  I think it's easier from a "labeled geometric diagram" to use these short symbol names but maybe harder when looking at the code for the first time?

renderlib/RotateTool.cpp
```diff
@@ -0,0 +1,367 @@
...
+  if (origins.empty()) {
+    origins.update(scene);
+  }
+  AffineSpace3f target = origins.currentReference(scene, Origins::kNormalize);
+
+  glm::vec3 viewDir = (scene.camera.m_From - target.p);
+  LinearSpace3f camFrame = scene.camera.getFrame();
+
+  // Draw the manipulator to be at some constant size on screen
+  float scale = length(viewDir) * scene.camera.getHalfHorizontalAperture() * (s_manipulatorSize / resolution.x);
```
ShrimpCryptid, line 232: Could you explain how this keeps the scale constant? I think I'm getting tripped up by the inclusion of the `getHalfHorizontalAperture()`, which is in radians.

renderlib/RotateTool.cpp
```diff
@@ -0,0 +1,367 @@
...
+    gesture.drawCone(axis.p,
+                     camFrame.vy * tumblescale, // swapped x and y to invert the triangles to face user
+                     camFrame.vx * tumblescale,
+                     glm::vec3(0, 0, 0),
+                     48,
+                     color,
+                     opacity,
+                     code);
+  }
+  static constexpr uint32_t noCode = -1;
```
ShrimpCryptid, line 257: Should this be moved to `Manipulator.h`?

renderlib/RotateTool.cpp
```diff
@@ -0,0 +1,367 @@
...
+    gesture.drawCircle(axis.p, axis.l.vz * axisscale, axis.l.vx * axisscale, 48, color, 1, code);
+  }
+  // Draw the z ring in xy plane
+  {
+    uint32_t code = manipulatorCode(RotateTool::kRotateZ, m_codesOffset);
+    glm::vec3 color = ManipColors::zAxis;
+    if (m_activeCode == RotateTool::kRotateZ) {
+      color = glm::vec3(1, 1, 0);
+      // if we are rotating, draw a tick mark where the rotation started, and where we are now
+      // Click in some proportional NDC: x [-1, 1] y [-aspect, aspect]
```
ShrimpCryptid, line 288: Remove?

renderlib/gesture/gesture.cpp
```diff
@@ -429,6 +429,32 @@ Gesture::Graphics::RenderBuffer::create(glm::ivec2 resolution, int samples)
...
+  // Render to texture
+  glBindFramebuffer(GL_FRAMEBUFFER, frameBuffer);
+  {
+    glViewport(0, 0, resolution.x, resolution.y);
+    uint32_t clearcode = Gesture::Graphics::SelectionBuffer::k_noSelectionCode;
+    glClearColor(((clearcode >> 0) & 0xFF) / 255.0,
+                 ((clearcode >> 8) & 0xFF) / 255.0,
+                 ((clearcode >> 16) & 0xFF) / 255.0,
+                 ((clearcode >> 24) & 0xFF) / 255.0);
+    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
```
ShrimpCryptid, line 450: Puzzling through this, do you need to specify both `GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT` because the frameBuffer uses color values? Where is depth being used?

renderlib/gesture/gesture.cpp
```diff
@@ -429,6 +429,32 @@ Gesture::Graphics::RenderBuffer::create(glm::ivec2 resolution, int samples)
...
+  }
+
+  // Restore
+  glBindFramebuffer(GL_FRAMEBUFFER, last_framebuffer);
+  glClearColor(last_clear_color[0], last_clear_color[1], last_clear_color[2], last_clear_color[3]);
+}
```
ShrimpCryptid, line 456: What does restoring the buffer here do?

ShrimpCryptid (approved):
Overall looks good! I had some questions mostly for my own knowledge.

renderlib/RotateTool.cpp
ShrimpCryptid: Reading through the rest of the codebase, I think this is fine. You explain `l + l0 + l1` everywhere they're used.

renderlib/gesture/gesture.cpp
toloudis: OpenGL is a state machine. glBindFramebuffer is setting the "current" framebuffer that OpenGL will draw to.  So restoring the previous one is the polite thing to do if the code doesn't want to know anything about who called it.

renderlib/gesture/gesture.cpp
toloudis: generally, when you create a framebuffer object you bind both a color buffer and a depth buffer to it. The pick buffer is no different - we want to use depth testing to allow pickable objects to occlude each other just as if we were rendering.  So it makes sense to clear the depth buffer before rendering to it.

ShrimpCryptid: Gotcha, thank you!

renderlib/RotateTool.cpp
toloudis: hm, yeah probably. There's actually a constant for it defined in the SelectionBuffer class. I should use that.

renderlib/RotateTool.cpp
toloudis: actually getHalfHorizontalAperture is not an angle.  It's half the horizontal width of the plane one unit away from the camera position, given a FoV angle.  (aka `tan(fov angle)` )

PR #145, opened 2023-10-06
Feature/transformable objects

toloudis:
This code change accomplishes the following:
1. begin to introduce an object hierarchy so that scene objects can be associated with transforms.  e.g. `SceneObject` is a container for a transform.  `SceneLight` is a `SceneObject` (which has a transform) that has a Light.
2. manipulators can operate on those transforms
3. fix several bugs with the rotation tool
4. split large files into separate files per class (AreaLightTool, MoveTool, RotateTool etc)
5. fix one floating-point shading bug on Windows in the gui shader
6. Area lights are currently manipulated by rotation of their direction vector only (as seen in `SceneLight::updateTransform`)


renderlib/Light.h
```diff
@@ -0,0 +1,115 @@
...
+    m_InvHeight = Other.m_InvHeight;
+    m_HalfHeight = Other.m_HalfHeight;
+    m_InvHalfHeight = Other.m_InvHalfHeight;
+    m_Distance = Other.m_Distance;
+    m_SkyRadius = Other.m_SkyRadius;
+    m_P = Other.m_P;
+    m_Target = Other.m_Target;
+    m_N = Other.m_N;
+    m_U = Other.m_U;
+    m_V = Other.m_V;
```
ShrimpCryptid, line 94: Do glm::vec3 values need to be copied? Not sure what their behavior is for equality operator

renderlib/Origins.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "Logging.h"
+#include "RenderSettings.h"
+
+void
+Origins::update(SceneView& scene)
+{
+  if (scene.scene) {
+
+    // e.g. find all selected objects in scene and collect up their centers/transforms here.
+    SceneLight& lt = scene.scene->m_lighting.m_sceneLights[1];
```
ShrimpCryptid, line 13: Is this a TODO?

renderlib/Light.h
toloudis: In c++ you can generally assume that assignment is a copy.

renderlib/Origins.cpp
toloudis: yes, once the framework is completely adopted / generalized, this code won't have to access the area light in this way - it will go through something like scene.selectedObject().

renderlib/AreaLightTool.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+  gesture.graphics.addCommand(GL_LINES);
+  gesture.graphics.addLine(Gesture::Graphics::VertsCode(p + v0 * 0.1f, color, opacity, code),
+                           Gesture::Graphics::VertsCode(t + v0 * 0.1f, color, opacity, code));
+  gesture.graphics.addLine(Gesture::Graphics::VertsCode(p + v1 * 0.1f, color, opacity, code),
+                           Gesture::Graphics::VertsCode(t + v1 * 0.1f, color, opacity, code));
+  gesture.graphics.addLine(Gesture::Graphics::VertsCode(p + v2 * 0.1f, color, opacity, code),
+                           Gesture::Graphics::VertsCode(t + v2 * 0.1f, color, opacity, code));
+  gesture.graphics.addLine(Gesture::Graphics::VertsCode(p + v3 * 0.1f, color, opacity, code),
+                           Gesture::Graphics::VertsCode(t + v3 * 0.1f, color, opacity, code));
```
ShrimpCryptid, line 51: This code is drawing lines from the light towards the target, right? Is there a reason for it not to be grouped with the existing line drawings above? (redundant call to `gesture.graphics.addCommand(GL_LINES)`?)

Also might be nice to add a comment for each of these blocks to say what they're drawing

renderlib/MathUtil.cpp
```diff
@@ -33,3 +33,11 @@ AffineSpace3f::inverse() const
...
   return AffineSpace3f(il, ip);
 }
+
+AffineSpace3f::AffineSpace3f(const glm::quat& orientation, const glm::vec3& p)
+  : p(p)
+{
+  // convert quat to linspace:
+  glm::mat3 m = glm::mat3_cast(orientation);
+  l = LinearSpace3f(m[0], m[1], m[2]);
+}
```
ShrimpCryptid, line 43: Does `l`/`LinearSpace3f` describe a rotation matrix?

renderlib/RotateTool.cpp
```diff
@@ -134,45 +134,45 @@ RotateTool::action(SceneView& scene, Gesture& gesture)
...
+        glm::vec3 vN = camFrame.vz;
+        angle = getDraggedAngle(camFrame.vz, p, l, l0, l1);
         motion = glm::angleAxis(angle, vN);
       } break;
       case RotateTool::kRotate: // general tumble rotation
         // use camera trackball algorithm
         // scale pixels to radians of rotation (TODO)
         float xRadians = -button.drag.x * dragScale * glm::two_pi<float>();
-        float yRadians = -button.drag.y * dragScale * glm::two_pi<float>();
+        float yRadians = button.drag.y * dragScale * glm::two_pi<float>();
```
ShrimpCryptid, line 175: There's another place in the code where a y-component also needs to be negated, is this related?

renderlib/RotateTool.cpp
```diff
@@ -227,13 +228,22 @@ RotateTool::draw(SceneView& scene, Gesture& gesture)
...
   // Draw the manipulator to be at some constant size on screen
   float scale = length(viewDir) * scene.camera.getHalfHorizontalAperture() * (s_manipulatorSize / resolution.x);
 
   AffineSpace3f axis;
   axis.p = target.p;
 
+  // TODO allow switching between local and world coordinate space!
+  // axis.l.vx = m_rotation * target.l.vx; // glm::vec3(1, 0, 0);
+  // axis.l.vy = m_rotation * target.l.vy; // glm::vec3(0, 1, 0);
+  // axis.l.vz = m_rotation * target.l.vz; // glm::vec3(0, 0, 1);
```
ShrimpCryptid, line 242: ooh that will be nice

renderlib/graphics/glsl/GLGuiShader.cpp
```diff
@@ -46,18 +46,19 @@ static const char* fragment_shader_text =
...
-            result.w = 1.0;
+        if (picking == 1) {
+          result.w = 1.0;
+        }
 
         // Gesture geometry handshake: any uv value below -64 means
         // no texture lookup. Check VertsCode::k_noTexture
-        if (picking == 0 && Frag_UV.s > -64)
-            result *= texture(Texture, Frag_UV.st);
+        if (picking == 0 && Frag_UV.x > -63.0) {
```
ShrimpCryptid, line 55: Should this be `Frag_UV.x >= -63`?

renderlib/Object3d.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "MathUtil.h"
+
+// an object that can be transformed in 3d space
+class Transform3d
+{
+public:
+  Transform3d()
+  {
+    m_rotation = glm::angleAxis(0.0f, glm::vec3(0.0f, 0.0f, 1.0f));
```
ShrimpCryptid, line 11: Is there a difference between using `angleAxis` and `glm::quat` here?

ShrimpCryptid (approved):
Looks good! Scattered some comments and questions in but nothing urgent.

renderlib/MathUtil.cpp
toloudis: yes, basically it's a 3x3 matrix (or set of 3 basis vectors) but it doesn't strictly have to be a rotation.  In all cases I'm using it, it is a rotation e.g. between camera and world space

renderlib/RotateTool.cpp
toloudis: Yeah - I'll look for it and see if they can be factored together

renderlib/RotateTool.cpp
toloudis: next PR will implement this

renderlib/graphics/glsl/GLGuiShader.cpp
toloudis: I'll add a comment here.  This was fixing a windows-only (and maybe nvidia-only) bug where the shader could not figure out whether `-64 > -64` .  The lines drawn were choppy.  So basically this is really Frag_UV.x >= (-64+epsilon)

renderlib/Object3d.h
toloudis: I'll switch it to the simpler version that initializes it with a vec3 of euler angles.

renderlib/CCamera.cpp
```diff
@@ -238,7 +208,11 @@ cameraManipulation(const glm::vec2 viewportSize, Gesture& gesture, CCamera& came
     glm::vec3 rotated_v = camera.m_From - camera.m_Target;
     glm::vec3 v = camera.m_From - camera.m_Target;
     glm::vec3 newEye, newUp;
-    trackball(drag.x, -drag.y, camera, newEye, newUp);
+
+    glm::quat q = trackball(drag.x, -drag.y, camera.m_From - camera.m_Target, camera.m_Up, camera.m_U);
+    newUp = q * camera.m_Up;
+    newEye = q * rotated_v + camera.m_Target;
+
     rotated_v = newEye - camera.m_Target;
```
ShrimpCryptid, line 216: Does adding and then removing `camera.m_Target` do anything here?

toloudis: nope! I wondered if anyone would notice 😁

toloudis: This is now resolved in one hopefully last commit on this PR

PR #147, opened 2023-10-18
Camera panel layout cleanup

toloudis:
Puts all the controls in the Camera panel into a single flat list, to be more consistent with the other UI control panels.  This is a small step toward UI uniformity / improvements.
Before:
agave_kkz4qaaqVc

After:
agave_t4aAUo7rVG


agave_app/FilmWidget.cpp
```diff
@@ -1,181 +0,0 @@
...
-    m_Preset[3].SetPreset(480, 384);
-  }
-
-  if (PresetType == "HDTV (video)") {
-    m_Preset[0].SetPreset(1920, 1080);
-    m_Preset[1].SetPreset(490, 270);
-    m_Preset[2].SetPreset(1280, 720);
-    m_Preset[3].SetPreset(320, 180);
-  }
-}
```
ShrimpCryptid, line 105: Just checking, is there still a way to change these presets? Were they removed because this all got moved to the Render dialog?

agave_app/CameraWidget.cpp
```diff
@@ -4,28 +4,154 @@
...
+  m_NoiseReduction.setCheckState(m_renderSettings->m_DenoiseParams.m_Enabled ? Qt::CheckState::Checked
+                                                                             : Qt::CheckState::Unchecked);
+  m_NoiseReduction.blockSignals(false);
+  emit m_qcamera->Changed();
+}
+void
+QCameraWidget::OnApertureChanged()
+{
+  m_ApertureSizeSlider.setValue(m_qcamera->GetAperture().GetSize(), true);
+  emit m_qcamera->Changed();
```
ShrimpCryptid, line 101: Do these need to be configured to block signals while being changed too?

ShrimpCryptid (approved):
Looks good! Love how much boilerplate code got stripped out by this change.

agave_app/FilmWidget.cpp
toloudis: Yeah this has been dead code since a long time ago. It is some last vestige from pre-1.0 agave.  The only render resolution settings are now in the Render Dialog.  The main viewport just always uses the window size, for now.

agave_app/CameraWidget.cpp
toloudis: the second argument to setValue is a boolean blockSignals parameter.  The emit m_qCamera0->Changed() is just a signal to say that the camera was updated but not via the UI.  This codepath is for receiving changes when a saved session is loaded from json, and we need to make sure the UI is updated.  So it allows two-way communication between the UI and QObjects holding state.

ascibisz (approved):

PR #149, opened 2023-10-20
allow dpr to factor into sizing of manipulators

toloudis:
Manipulators were showing up with bad scaling on screens with different device pixel ratios.  

This code change applies device pixel ratio to manipulator sizing, and tries to re-set it on window resize in case the window has been dragged from one screen to another.
The dpr comes from Qt code, so it is accessed in the gui layer and then premultiplied before being passed down to the manipulators themselves.


frasercl (approved):

ShrimpCryptid (approved):
Looks good!

Issue #142, opened 2023-10-03
help / resources


toloudis:
help menu / about box,  links to online docs, credits/copyright/license,  any help resources like tutorials

TravisKroeker:
This ticket is in part outcome from a comprehensive User Support Discovery. [Link to Miro Board with findings here](https://miro.com/app/board/uXjVMpyjeUs=/?share_link_id=632071778389)

**Summary:**
Add Help section to main toolbar (preliminary designs completed). Potentially includes:
- About section?
- Link to Forum (labeled Questions?)
- Link to AGAVE website (Learn More About AGAVE...)
- Citation Guidance (user request from Forum... in Ux backlog for awhile). Could be direct link to guidance on website or a modal of some sort. Ability to click one button to copy citation?
- Documentation
- Walkthrough tutorial (eventually)

toloudis:
@TravisKroeker the miro board link is not letting me in. Can you unprotect it?

TravisKroeker:
I think I changed the permissions to allow you access but it looks you completed the task anyway.

From: toloudis ***@***.***>
Date: Wednesday, December 13, 2023 at 1:38 PM
To: allen-cell-animated/agave ***@***.***>
Cc: Travis Kroeker ***@***.***>, Mention ***@***.***>
Subject: Re: [allen-cell-animated/agave] help / resources (Issue #142)

@TravisKroeker<https://github.com/TravisKroeker> the miro board link is not letting me in. Can you unprotect it?

toloudis:
@TravisKroeker Yes I found the other board with the actual help dialog design on it, and completed the first version of the Help menu and dialogs.  Any polish/details remaining can go in fresh tickets here.

PR #154, opened 2023-11-22
Help menu and About dialog

toloudis:
Initial add of a help menu and about dialog, to get us going.  This addresses most if not all of #142 

Help menu consists of 
* link to image.sc forum's agave tag
* link to agave issues board to report bugs
* link to agave documentation
* link to main agave github page

The about box also contains some information about how to cite AGAVE in publications.

There are still some possible to-do's here such as license information (or links to it).

Screenshot from macOS: 

About AGAVE 2023-11-21 16-32-07

### DIFF
--- a/agave_app/aboutDialog.cpp
+++ b/agave_app/aboutDialog.cpp
@@ -14,7 +14,7 @@ AboutDialog::AboutDialog()
   auto label = new QLabel(this);
 
   label->setText("<h1><b>AGAVE</b></h1>"
-                 "<h4><i>Advanced GPU Accelerated Volume Explorer</i></h4>"
+                 "<i>Advanced GPU Accelerated Volume Explorer</i>"
                  "<h4><b>Version " +
                  qApp->applicationVersion() +
                  "</b></h4>"
--- a/agave_app/aboutDialog.cpp
+++ b/agave_app/aboutDialog.cpp
@@ -13,8 +13,8 @@ AboutDialog::AboutDialog()
   auto layout = new QVBoxLayout(this);
   auto label = new QLabel(this);
 
-  label->setText("<h1><b>AGAVE</b></h1>"
-                 "<i>Advanced GPU Accelerated Volume Explorer</i>"
+  label->setText("<h1 style=\"line-height:70%\"><b>AGAVE</b></h1>"
+                 "<i  style=\"line-height:70%\">Advanced GPU Accelerated Volume Explorer</i>"
                  "<h4><b>Version " +
                  qApp->applicationVersion() + "</b></h4>");
   label->setAlignment(Qt::AlignCenter);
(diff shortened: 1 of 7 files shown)


agave_app/aboutDialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+  label->setAlignment(Qt::AlignCenter);
+  layout->addWidget(label);
+
+  label = new QLabel(this);
+  label->setText(
+    "<a href=\"https://forum.image.sc/tag/agave\">Support forum</a><br/>"
+    "<a href=\"https://allen-cell-animated.github.io/agave\">Online documentation</a><br/>"
+    "<a href=\"https://github.com/allen-cell-animated/agave/issues\">Report a bug or make a request</a><br/>"
+    "<a href=\"https://github.com/allen-cell-animated/agave\">Source code</a><br/>"
+    "");
```
ShrimpCryptid, line 31: Nit: If we ever add more links it might be nice to use a shared constants file across this and the other actions down below.

ShrimpCryptid (commented):
image

Here's how it looks on Windows, also! Everything seems to be working (I clicked through all the links without issues).

agave_app/aboutDialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+    "If you use AGAVE in your research, please cite the following:<br>"
+    "<br/>"
+    "Daniel Toloudis, AGAVE Contributors (2023). AGAVE: Advanced GPU Accelerated Volume Explorer (Version " +
+    qApp->applicationVersion() +
+    ")"
+    "[Computer software]. Allen Institute for Cell Science. <a "
+    "href=\"" +
+    agaveUrl +
+    "\"></a><br/>"
+    "<br/>"
```
ShrimpCryptid, line 53: Nit: I think it might be a little easier to use the citations if we put the citation information (both this and the bibtex info) in their own `QTextBrowser` block, and hoist the rest of the about text out to be on the same level as the links? That way, users can double click to copy the full citation information within the blocks.

ShrimpCryptid (approved):
Small nits, otherwise looks good!

agave_app/aboutDialog.cpp
toloudis: I redid all this after seeing Travis's idea to put citation info in a separate dialog.  Citation is now a help menu item.

toloudis:
After some tweaks, here are the new dialogs on macos:

About AGAVE 2023-12-13 13-40-33
Cite AGAVE 2023-12-13 13-40-49

agave_app/aboutDialog.cpp
toloudis: removed from "about" dialog

PR #158, opened 2024-01-12
Feature/scalebar

toloudis:
This PR adds a visual scale bar in the lower right corner of the image.
In perspective projection, because of foreshortening, the scale bar refers to the distance between tick-marks displayed on the bounding box.  
In orthographic projection, the scale bar is correctly proportional for all camera angles since there is no foreshortening for orthographic projections.

Most important files to look at:
* ScaleBarTool.cpp (draws the actual scale bar)
* Font.cpp (encapsulate a texture map with a rendered ttf font)
* gesture.cpp (how to draw text as one aligned bitmap quad per character)
* Util.cpp (how to draw tickmarks in 3d)

Nearly all the rest is "plumbing".

Code changes:
* add handling of units to all file i/o.  The reported VolumeDimensions now have a units string.  There is no intelligence about units at all - they are just treated as plain strings.
* add ScaleBarTool. responsible for calculating the size and drawing the scale bar
* add code to render tickmarks along the bounding box
* add a bitmapped font generator so that text can be easily displayed on the viewport without overlays .  For simplicity, no special characters are allowed in units strings, so there is a small utility function to clean a couple of known unit characters. (angstrom and micron)
* add stb_image and stb_font third party code for ttf font handling and image loading. No need to review the stb files; they are simply copy-pasted in.
* add build infrastructure to distribute the .ttf file as a bundled data file and runtime code to find and load it at runtime
* ensure that rendering is consistent between interactive viewport and rendered image
* add scale bar support to python api, web api, and serialization

Perspective mode:
AGAVE 1 5 0 2024-01-16 13-10-13
Orthographic mode
AGAVE 1 5 0 2024-01-16 13-10-42


renderlib/graphics/stb/stb_image.h
toloudis, line 1: ignore this file for review: third party code

renderlib/graphics/stb/stb_truetype.h
toloudis, line 1: ignore this file for review: third party code

agave_app/CMakeLists.txt
```diff
@@ -99,9 +99,52 @@ target_link_libraries(agaveapp PRIVATE
...
+	# Individually set the file's path properties.
+	foreach (FILE ${RESOURCES})
+		# Get the relative path from the data-folder to the particular file.
+		file(RELATIVE_PATH NEW_FILE "${RENDERLIB_SOURCE_DIR}" ${FILE})
 
-	# RUNTIME_DEPENDENCIES DIRECTORIES $ENV{Qt6_DIR}/lib
+		# Get the relative path to the file.
+		get_filename_component(NEW_FILE_PATH ${NEW_FILE} DIRECTORY)
+
+		# Set it's location inside the app package (under Resources).
```
ShrimpCryptid, line 132:
```suggestion
		# Set its location inside the app package (under Resources).

```

agave_app/CMakeLists.txt
```diff
@@ -99,9 +99,53 @@ target_link_libraries(agaveapp PRIVATE
...
+	add_custom_command(
+		TARGET agaveapp POST_BUILD
+		COMMAND ${CMAKE_COMMAND} -E copy_directory
+			"$<TARGET_PROPERTY:renderlib,SOURCE_DIR>/assets"
+			"$<TARGET_PROPERTY:agaveapp,BINARY_DIR>/assets"
+	)
+elseif(APPLE)
+	get_target_property(RENDERLIB_SOURCE_DIR renderlib SOURCE_DIR)
+	# Get a list of all of the files in the data-folder.
+	# file(GLOB_RECURSE RESOURCES "$<TARGET_PROPERTY:renderlib,SOURCE_DIR>/assets/*")
```
ShrimpCryptid, line 114: Remove?

agave_app/agaveGui.cpp
```diff
@@ -483,6 +481,9 @@ agaveGui::onRenderAction()
 
   // copy of camera
   CCamera camera = m_glView->getCamera();
+  // extract ViewerWindow from GLView3D to hand to RenderDialog
+  ViewerWindow* renderer = m_glView->borrowRenderer();
```
ShrimpCryptid, line 485: What did this change to `ViewerWindow` from `IRenderWindow` do?

renderlib/VolumeDimensions.cpp
```diff
@@ -187,6 +188,15 @@ VolumeDimensions::getChannelNames(const std::vector<uint32_t>& channels) const
...
 
+std::string
+VolumeDimensions::sanitizeUnitsString(std::string units)
+{
+  // sanitize known weird characters
+  units = std::regex_replace(units, std::regex("µ"), "u");
+  units = std::regex_replace(units, std::regex("Å"), "A");
+  return units;
+}
+
```
ShrimpCryptid, line 199: Do the font files not include these characters?

renderlib/io/FileReaderCCP4.cpp
```diff
@@ -165,16 +169,78 @@ readCCP4Dimensions(const std::string filepath, VolumeDimensions& dims, uint32_t
...
+  if (!myFile || myFile.gcount() != 4) {
+    LOG_ERROR << "Bad file read from " << filepath;
+    return false;
+  }
+  // NZ              Number of intervals along Z
+  myFile.read((char*)&dummy, 4);
+  if (!myFile || myFile.gcount() != 4) {
+    LOG_ERROR << "Bad file read from " << filepath;
+    return false;
+  }
```
ShrimpCryptid, line 216: A lot of repeated code here it looks like... can this be a for loop?

agave_app/agaveGui.cpp
toloudis: The idea here was that the RenderDialog might need to use more things from the GLView than just its Renderer.  
It used to be GLView-->(has a ) IRenderWindow,  but now GLView has-a ViewerWindow has-a IRenderWindow.

toloudis: There may soon be a Great Renaming of these windows and renderer things

renderlib/VolumeDimensions.cpp
toloudis: Yes - the generated fonts are trying to capture a minimal set of characters to save texture space.   Ithink the API wants you to only use one contiguous range of character codes.  Maybe in a future PR I will figure out how to keep the nice µ character.

renderlib/io/FileReaderCCP4.cpp
toloudis: Yeah I could probably move this into a function or sometihng.  Or maybe just keep track of the data instead of using dummy vars.

renderlib/ScaleBarTool.cpp
```diff
@@ -0,0 +1,125 @@
...
+{
+  // note this result will always be some integer power of 10 independent of zoom...
+  return pow(10.0f, floor(log10(physicalScale / 2.0f)));
+}
+
+static float
+getScaleBarWidthOrtho(int windowWidthPx, float windowPhysicalWidth)
+{
+  // We want to find the largest round number of physical units that keeps the scale bar within this width on screen
+  const int SCALE_BAR_MAX_WIDTH = (int)((float)windowWidthPx * 0.25f);
```
ShrimpCryptid, line 21: Should this be in all-caps notation if it's just a local var?

Also nit, might be good to include units at the end of this variable name (`scaleBarMaxWidthPx`)

renderlib/ScaleBarTool.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include <iomanip>
+#include <sstream>
+
+// physicalscale is max of physical dims x,y,z
+static float
+computePhysicalScaleBarSize(const float physicalScale)
+{
+  // note this result will always be some integer power of 10 independent of zoom...
+  return pow(10.0f, floor(log10(physicalScale / 2.0f)));
+}
```
ShrimpCryptid, line 15: What does dividing by 2 here do?

renderlib/ScaleBarTool.cpp
ShrimpCryptid, line 1: Were you able to validate the resulting scale bar to make sure the displayed units were accurate?

renderlib/gesture/gesture.cpp
```diff
@@ -657,3 +659,54 @@ Gesture::drawCone(glm::vec3 base,
...
+  // attach the texture id with the draw command.
+  graphics.glTextureId = graphics.font->getTextureID();
+
+  // assume orthographic projection with units = screen pixels, origin at top left
+  // also assume we are in a "TRIANGLES" draw command.
+
+  stbtt_aligned_quad q;
+  const char* text = stext.c_str();
+  float lastxpos = xpos;
+  float lastypos = ypos;
```
ShrimpCryptid, line 681: Unused?

renderlib/gesture/gesture.cpp
```diff
@@ -657,3 +659,54 @@ Gesture::drawCone(glm::vec3 base,
...
+      graphics.addVert(
+        Gesture::Graphics::VertsCode(glm::vec3(q.x1, q.y1, 0.0), glm::vec2(q.s1, q.t1), color, opacity, code));
+      // 0
+      graphics.addVert(
+        Gesture::Graphics::VertsCode(glm::vec3(q.x0, q.y0, 0.0), glm::vec2(q.s0, q.t0), color, opacity, code));
+      // 3
+      graphics.addVert(
+        Gesture::Graphics::VertsCode(glm::vec3(q.x0, q.y1, 0.0), glm::vec2(q.s0, q.t1), color, opacity, code));
+    }
+    ++text;
```
ShrimpCryptid, line 710: Low-level question, how does drawing the two triangles of the quad result in the quad's text being drawn? Is `getBakedQuad` setting OpenGL to use the text's texture somewhere?

renderlib/ViewerWindow.cpp
```diff
@@ -171,6 +174,13 @@ ViewerWindow::redraw()
     LOG_ERROR << "Failed to update selection buffer";
   }
 
+  // lazy init
+  if (!gesture.graphics.font.get()) {
+    gesture.graphics.font.reset(new Font());
+    std::string fontPath = renderlib::assetPath() + "/fonts/Arial.ttf";
+    gesture.graphics.font->load(fontPath.c_str());
+  }
+
```
ShrimpCryptid, line 183: `renderer.cpp`/`pyrenderer.cpp` also have this same initialization logic. Does this need to be in both places?

renderlib/graphics/gl/Font.cpp
```diff
@@ -0,0 +1,146 @@
...
+generateFontTexture(unsigned char* temp_bitmap)
+{
+  // expand temp_bitmap to RGBA
+  unsigned char* expanded_bitmap = new unsigned char[s_textureSize * s_textureSize * 4];
+  for (int i = 0; i < s_textureSize * s_textureSize; ++i) {
+    expanded_bitmap[i * 4 + 0] = 255;
+    expanded_bitmap[i * 4 + 1] = 255;
+    expanded_bitmap[i * 4 + 2] = 255;
+    expanded_bitmap[i * 4 + 3] = temp_bitmap[i];
+  }
```
ShrimpCryptid, line 22: Oh interesting, is the font texture only stored as an alpha/greyscale value?

renderlib/graphics/gl/Font.cpp
```diff
@@ -0,0 +1,146 @@
...
+  unsigned char temp_bitmap[s_textureSize * s_textureSize];
+  stbtt_BakeFontBitmap(buffer,
+                       0,
+                       32.0,
+                       temp_bitmap,
+                       s_textureSize,
+                       s_textureSize,
+                       m_firstChar,
+                       m_numChars,
+                       m_cdata); // no guarantee this fits!
```
ShrimpCryptid, line 75: Scary... If we ever do run over this texture size limit, how will we be able to tell? Will it just cut off out-of-bounds characters or will it start to rewrite parts of the texture array?

renderlib/graphics/gl/Util.cpp
```diff
@@ -337,6 +354,113 @@ BoundingBoxDrawable::drawLines(const glm::mat4& transform, const glm::vec4& colo
   _shader->release();
 }
 
+// physicalscale is max of physical dims x,y,z
+static float
+computePhysicalScaleBarSize(const float physicalScale)
+{
+  // note this result will always be some integer power of 10 independent of zoom...
+  return pow(10.0f, floor(log10(physicalScale / 2.0f)));
+}
```
ShrimpCryptid, line 363: Copied in `ScaleBarTool.cpp`?

renderlib/ScaleBarTool.cpp
toloudis: Thanks, yeah. It used to be a true constant but not anymore. Will fix.

renderlib/ScaleBarTool.cpp
toloudis: Yes but I only eyeballed it on OME-TIFFs and one Zarr. In Agave there is the statistics view which shows the volume physical size, and you can verify that the scalebar is about the right proportion.  I switched ortho and perspective and zoomed in and out.  Before release I will need to do more testing with CZI files too.

toloudis: The units themselves were validated similarly but admittedly I don't have a wide variety of units available to me in test data.

renderlib/ViewerWindow.cpp
toloudis: Yeah for now I have stuck this font as a member of Gesture.Graphics so each separate instance of Graphics needs to do its own font init.  I likely will move it to be owned by `renderlib` so it's more of a global singleton.

renderlib/graphics/gl/Font.cpp
toloudis: yep. Shader can apply color if it's ever needed.

renderlib/graphics/gl/Util.cpp
toloudis: - [ ] refactor this so it's not duplicated

renderlib/gesture/gesture.cpp
toloudis: several lines above this while loop, in this same function, I set the active texture to be the one from the font.

toloudis: and the s,t coordinates are the position of the character glyph on the font texture

renderlib/ScaleBarTool.cpp
toloudis: good question.  I think I saw it in volume viewer and thought it was just giving some rounding-down effect to the result value.  I'll test with and without

renderlib/io/FileReaderCCP4.cpp
toloudis: Opting to leave this as-is for later when I will very likely read the values into actual variables to be used.

renderlib/graphics/gl/Font.cpp
toloudis: I think this is one of those things that is boilerplate example code from the font lib I am using, and it was more effort to properly calculate this than the risk of it actually going bad in the general use case.  I'll write a new ticket just to keep it on the table.

ShrimpCryptid (approved):
Tested locally and things seem to be working just fine! Thanks for addressing my feedback.

PR #159, opened 2024-01-16
Feature/light manip UI

toloudis:
A key missing component of in-viewport light manipulator.   User interaction with the manipulator was not updating the numeric entry fields in the user interface.  This change resolves that.

It also adds some code to the keypress handler so that the rotation gizmo and the light gizmo are in sync.  
There is a button in the area light control panel to toggle the in-viewport controls. Or press "R" to toggle light source rotation gizmo on/off.   
There is also a translation gizmo but since it is non-functional, it will stay disconnected from the UI.

### DIFF
--- a/agave_app/GLView3D.cpp
+++ b/agave_app/GLView3D.cpp
@@ -287,10 +287,46 @@ GLView3D::FitToScene()
 }
 
 void
+GLView3D::toggleAreaLightRotateControls()
+{
+  // toggle rotate tool
+  if (m_areaLightMode == AREALIGHT_MODE::NONE || m_areaLightMode == AREALIGHT_MODE::TRANS) {
+    m_viewerWindow->showAreaLightGizmo(true);
+    m_viewerWindow->setTool(
+      new RotateTool(m_viewerWindow->m_toolsUseLocalSpace, ManipulationTool::s_manipulatorSize * devicePixelRatioF()));
+    m_viewerWindow->forEachTool(
+      [this](ManipulationTool* tool) { tool->setUseLocalSpace(m_viewerWindow->m_toolsUseLocalSpace); });
+    m_areaLightMode = AREALIGHT_MODE::ROT;
+  } else {
+    m_viewerWindow->showAreaLightGizmo(false);
+    m_viewerWindow->setTool(nullptr);
+    m_areaLightMode = AREALIGHT_MODE::NONE;
+  }
+}
+
+// TODO currently this function is not wired up to any gui at all.
+// This is because translation of area light source still needs work.
+// (Currently rotation is sufficient.)
+void
+GLView3D::toggleAreaLightTranslateControls()
+{
+  // toggle translate tool
+  if (m_areaLightMode == AREALIGHT_MODE::NONE || m_areaLightMode == AREALIGHT_MODE::ROT) {
+    m_viewerWindow->showAreaLightGizmo(true);
+    m_viewerWindow->setTool(
+      new MoveTool(m_viewerWindow->m_toolsUseLocalSpace, ManipulationTool::s_manipulatorSize * devicePixelRatioF()));
+    m_viewerWindow->forEachTool(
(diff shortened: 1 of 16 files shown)


renderlib/ViewerWindow.cpp
```diff
@@ -20,16 +20,17 @@ ViewerWindow::ViewerWindow(RenderSettings* rs)
   // TODO have a notion of a scene's selection set,
   // and activate tools via the UI to operate
   // on the selection set.
-  // TEST create a tool and activate it
-  // m_activeTool should not be in m_tools
-  // m_activeTool = new MoveTool();
-  // m_activeTool = new RotateTool();
-  // m_tools.push_back(new AreaLightTool());
+  m_areaLightTool = new AreaLightTool();
```
toloudis, line 23: This code is showing two different patterns: one where I abstractly add a "disableable" tool to the list of tools (new ScaleBarTool below), and one where I explicitly create a tool and expect outside code to add/remove it as needed.  I'm not sure if future work in this code will unify the two approaches, but both work for now as long as the bookkeeping is done properly.

agave_app/GLView3D.cpp
```diff
@@ -287,41 +287,53 @@ GLView3D::FitToScene()
...
+      new MoveTool(m_viewerWindow->m_toolsUseLocalSpace, ManipulationTool::s_manipulatorSize * devicePixelRatioF()));
+    m_viewerWindow->forEachTool(
+      [this](ManipulationTool* tool) { tool->setUseLocalSpace(m_viewerWindow->m_toolsUseLocalSpace); });
+    m_areaLightMode = AREALIGHT_MODE::TRANS;
+  } else {
+    m_viewerWindow->showAreaLightGizmo(false);
+    m_viewerWindow->setTool(nullptr);
+    m_areaLightMode = AREALIGHT_MODE::NONE;
+  }
+}
```
ShrimpCryptid, line 326: A note for the future, there's some duplicated boilerplate between these two methods. Might be worth seeing if they can be combined or reduced whenever the translation controls go in.

renderlib/ViewerWindow.cpp
```diff
@@ -50,6 +51,21 @@ ViewerWindow::setSize(int width, int height)
...
+{
+  // check if m_areaLightTool is in m_tools
+  bool isShowing = std::find(m_tools.begin(), m_tools.end(), m_areaLightTool) != m_tools.end();
+  if (isShowing == show) {
+    return;
+  }
+  if (show) {
+    m_tools.push_back(m_areaLightTool);
+  } else {
+    m_tools.erase(std::remove(m_tools.begin(), m_tools.end(), m_areaLightTool), m_tools.end());
```
ShrimpCryptid, line 65: If there are other tools in `m_tools` after `m_areaLightTool`, would this potentially erase/hide other tools unintentionally?

ShrimpCryptid (commented):
Overall looks good to me, built and tested locally without any issues. I left a couple questions about potential bugs and I'll approve once those are resolved.

renderlib/ViewerWindow.cpp
toloudis: This is sort of a standard recipe for "get rid of one element from a vector, by value".  std::remove finds the thing and makes sure it is shifted to the end of the list, and returns an iterator to the start of the removed things... and vector::erase drops it from m_tools.

agave_app/GLView3D.cpp
toloudis: Yes, that's something for later. There's a big TODO at the top of the translate function and it may be completely changed when it's time to wire it up to something.  This code is here to sort of remind us that it's similar to the rotate code but they can work together with the mode variable. Think I should just delete it?

ShrimpCryptid: I think it's fine to keep in, it's a good reminder.

ShrimpCryptid (approved):

PR #160, opened 2024-01-17
add option to keep current agave settings when loading

toloudis:
If loading a new volume, offer a LoadDialog checkbox to keep or discard current volume appearance settings.  Default is to keep settings.   This is to make it efficient when loading consecutive images of same channels and approximate intensity ranges.  Unchecking the box will present the previous behavior, where every setting goes back to a default.

Example: load a 9-channel variance dataset volume, enable a set of channels. Customize channel colors.  Adjust thresholding.  Then load another 9-channel variance dataset volume.  Same channels should be enabled, with same colors etc.

### DIFF
--- a/agave_app/agaveGui.cpp
+++ b/agave_app/agaveGui.cpp
@@ -543,7 +543,9 @@ agaveGui::onImageLoaded(std::shared_ptr<ImageXYZC> image,
                         const LoadSpec& loadSpec,
                         uint32_t sizeT,
                         const Serialize::ViewerState* vs,
-                        std::shared_ptr<IFileReader> reader)
+                        std::shared_ptr<IFileReader> reader,
+                        // only used if vs is null
+                        bool keepCurrentUISettings)
 {
   m_loadSpec = loadSpec;
 
@@ -564,12 +566,17 @@ agaveGui::onImageLoaded(std::shared_ptr<ImageXYZC> image,
   // Show timeline widget if the loaded image has multiple frames.
   m_timelinedock->setVisible(sizeT > 1);
 
+  bool wasVolumeLoaded = m_appScene.m_volume != nullptr;
+
   // install the new volume image into the scene.
   // this is deref'ing the previous _volume shared_ptr.
   m_appScene.m_volume = image;
 
-  m_appScene.initSceneFromImg(image);
-  m_glView->initCameraFromImage(&m_appScene);
+  m_appScene.initBoundsFromImg(image);
+  if (!keepCurrentUISettings || !wasVolumeLoaded) {
+    m_appScene.initSceneFromImg(image);
+    m_glView->initCameraFromImage(&m_appScene);
+  }
 
   // initialize _appScene from ViewerState
   if (vs) {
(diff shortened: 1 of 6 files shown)


agave_app/agaveGui.cpp
```diff
@@ -564,12 +566,17 @@ agaveGui::onImageLoaded(std::shared_ptr<ImageXYZC> image,
...
+  bool wasVolumeLoaded = m_appScene.m_volume != nullptr;
+
   // install the new volume image into the scene.
   // this is deref'ing the previous _volume shared_ptr.
   m_appScene.m_volume = image;
 
-  m_appScene.initSceneFromImg(image);
-  m_glView->initCameraFromImage(&m_appScene);
+  m_appScene.initBoundsFromImg(image);
+  if (!keepCurrentUISettings || !wasVolumeLoaded) {
```
toloudis, line 576: this is where we use the checkbox value to either reset settings or not

agave_app/loadDialog.cpp
```diff
@@ -172,6 +175,8 @@ LoadDialog::LoadDialog(std::string path, const std::vector<MultiscaleDims>& dims
   layout->addItem(new QSpacerItem(0, spacing, QSizePolicy::Expanding, QSizePolicy::Expanding));
   layout->addRow(mMemoryEstimateLabel);
   layout->addItem(new QSpacerItem(0, spacing, QSizePolicy::Expanding, QSizePolicy::Expanding));
+  layout->addRow("Keep current AGAVE settings", m_keepSettingsCheckbox);
+  layout->addItem(new QSpacerItem(0, spacing, QSizePolicy::Expanding, QSizePolicy::Expanding));
```
ShrimpCryptid, line 179: Nit: AGAVE settings to me suggests the configuration/settings for the entire AGAVE application (I don't know if we actually have saved preferences, now that I think about it), rather than just the current viewer settings. Would "viewer settings" or "volume settings" be more clear?

ShrimpCryptid:
It seems like any changes to the ROI on the LoadDialog is overridden when keeping the existing settings. Is this intended behavior?

UI has an existing ROI, and the LoadDialog has a different ROI set:
image

Once the image is loaded, it keeps the old ROI range, ignoring what was set in LoadDialog:
image

ShrimpCryptid (approved):
LGTM! Built and tested locally.

toloudis:
> It seems like any changes to the ROI on the LoadDialog are overridden when keeping the existing settings. Is this intended behavior?
> 
> UI has an existing ROI, and the LoadDialog has a different ROI set... 
> Once the image is loaded, it keeps the old ROI range, ignoring what was set in LoadDialog...

I think this is actually the correct/intended behavior.  The ROI in the loading dialog is letting you select a subset of what's in the original file to load.  Once that is loaded, that volume becomes the ENTIRE volume that the app knows about.   So if your loading ROI starts at x0, y0, z0 and spans a size of dx, dy, dz, then the LOADED volume will start at 0,0,0 and be of size dx,dy,dz.  

The saved settings of ROI in the Appearance panel preserve the relative ROI selection relative to the volume that was loaded.

ShrimpCryptid (approved):
Oops, thought I hadn't approved yet and resubmitted

PR #165, opened 2024-02-23
UI layout update

toloudis:
another pre-release 1.6 polish update.

* update Qt LTS version from 6.5.1 to 6.5.3 to pick up incidental bug fixes
* set initial app size to about 80% of screen size when launching.
* use Qt "Fusion" theme for consistent look across Windows/Mac/Linux (see https://doc.qt.io/qt-6/gallery.html)
* add small workaround for macos bug fix with layouts that include a QPushButton
* separate Area Light and Sky Light into their own sections
* adjust number of decimals in a couple of key places where they didn't make sense

Estimated time to review: 15min or less


agave_app/Controls.h
```diff
@@ -191,22 +193,25 @@ private slots:
   QSlider m_slider;
 };
 
-class Controls
+class AgaveFormLayout : public QGridLayout
```
toloudis, line 196: introducing this class allowed finer control over laying out collections of input controls and lets us ensure some consistency across the ui.

agave_app/Controls.cpp
```diff
@@ -1,18 +1,21 @@
...
 QColorPushButton::QColorPushButton(QWidget* pParent)
   : QPushButton(pParent)
-  , m_Margin(5)
+  , m_Margin(4)
   , m_Radius(4)
   , m_Color(Qt::gray)
 {
   setText("");
+  // deal with a macos style bug which causes this button to misalign in some situations
+  setAttribute(Qt::WA_LayoutUsesWidgetRect);
```
ShrimpCryptid, line 18: Is this still a bug with the fusion style?

ShrimpCryptid (approved):
LGTM, thank you for the occasional comments explaining your changes!

agave_app/Controls.cpp
toloudis: It's not a problem with fusion, but it still doesn't hurt to leave it in.  If we ever switch back it would suddenly break something.

agave_app/AppearanceSettingsWidget.cpp
```diff
@@ -260,17 +262,17 @@ QAppearanceSettingsWidget::createLightingControls(QAction* pLightRotationAction)
...
-  sectionLayout->addRow("AreaLight Size", m_lt0gui.m_sizeSlider);
+  sectionLayout->addRow("Size", m_lt0gui.m_sizeSlider);
   QObject::connect(
     m_lt0gui.m_sizeSlider, &QNumericSlider::valueChanged, this, &QAppearanceSettingsWidget::OnSetAreaLightSize);
 
   m_lt0gui.m_distSlider = new QNumericSlider();
   m_lt0gui.m_distSlider->setStatusTip(tr("Set distance for area light"));
   m_lt0gui.m_distSlider->setToolTip(tr("Set distance for area light"));
-  m_lt0gui.m_distSlider->setRange(0.1, 100.0);
+  m_lt0gui.m_distSlider->setRange(0.1, 10.0);
```
ascibisz, line 272: Out of curiosity, why did you change this value from 100 to 10?

toloudis: This value is a distance from light source to volume.  The volume size is normalized to 1.0 so putting the light 100 units away is not very useful.  It keeps the slider range more manageable for practical numbers.  The good news is that modifying the slider range does not hurt the ability to load and save values outside this range.

ascibisz (approved):

PR #169, opened 2024-03-18
Rotate tool should only show front facing rings

toloudis:
The rotate tool was drawing complete axis aligned rings which could create confusing interactions because you can't tell which part of the ring is facing forward.
This code change only shows the part of the ring that is forward facing and only allows drag interactions with that part of the ring.  This makes axis constrained drag rotation much more intuitive.

Before:
AGAVE 1 5 0 2024-03-18 11-41-34
After:
AGAVE 1 5 0 2024-03-18 11-40-26

### DIFF
--- a/renderlib/CCamera.cpp
+++ b/renderlib/CCamera.cpp
@@ -126,6 +126,7 @@ cameraManipulationDolly(const glm::vec2 viewportSize,
     camera.m_From += motion;
     camera.m_Target += targetMotion;
     camera.m_OrthoScale *= factor;
+    camera.Update();
 
     // Consume gesture on button release
     Gesture::Input::reset(button);
--- a/renderlib/RotateTool.cpp
+++ b/renderlib/RotateTool.cpp
@@ -261,6 +261,8 @@ RotateTool::draw(SceneView& scene, Gesture& gesture)
 
   gesture.graphics.addCommand(GL_LINES);
 
+  glm::vec4 discPlane(camFrame.vz, -glm::dot(camFrame.vz, axis.p));
+
   float axisscale = scale * 0.85f;
   // Draw the x ring in yz plane
   {
@@ -269,7 +271,7 @@ RotateTool::draw(SceneView& scene, Gesture& gesture)
     if (m_activeCode == RotateTool::kRotateX) {
       color = glm::vec3(1, 1, 0);
     }
-    gesture.drawCircle(axis.p, axis.l.vy * axisscale, axis.l.vz * axisscale, 48, color, 1, code);
+    gesture.drawCircle(axis.p, axis.l.vy * axisscale, axis.l.vz * axisscale, 48, color, 1, code, &discPlane);
   }
   // Draw the y ring in xz plane
   {
@@ -278,7 +280,7 @@ RotateTool::draw(SceneView& scene, Gesture& gesture)
     if (m_activeCode == RotateTool::kRotateY) {
       color = glm::vec3(1, 1, 0);
     }
-    gesture.drawCircle(axis.p, axis.l.vz * axisscale, axis.l.vx * axisscale, 48, color, 1, code);
+    gesture.drawCircle(axis.p, axis.l.vz * axisscale, axis.l.vx * axisscale, 48, color, 1, code, &discPlane);
   }
   // Draw the z ring in xy plane
   {
@@ -287,7 +289,7 @@ RotateTool::draw(SceneView& scene, Gesture& gesture)
     if (m_activeCode == RotateTool::kRotateZ) {
       color = glm::vec3(1, 1, 0);
     }
-    gesture.drawCircle(axis.p, axis.l.vx * axisscale, axis.l.vy * axisscale, 48, color, 1, code);
+    gesture.drawCircle(axis.p, axis.l.vx * axisscale, axis.l.vy * axisscale, 48, color, 1, code, &discPlane);
   }
 
   // Draw the camera-axis rotation manipulator as a circle always facing the view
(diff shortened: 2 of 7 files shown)


renderlib/RotateTool.cpp
```diff
@@ -261,6 +261,8 @@ RotateTool::draw(SceneView& scene, Gesture& gesture)
 
   gesture.graphics.addCommand(GL_LINES);
 
+  glm::vec4 discPlane(camFrame.vz, -glm::dot(camFrame.vz, axis.p));
+
```
ShrimpCryptid, line 265: Could you break down what this line is doing? I'm assuming it's making a normal vector pointing towards the camera (to be the clipping plane's normal vector) but I'm not quite following how it's getting there.

Is `camFrame.vz` the vector pointing out/away from the camera, and the dot product is getting a magnitude along that vector...?

ShrimpCryptid (approved):

renderlib/RotateTool.cpp
toloudis: This is standard geometry/linear algebra. The construct for a plane: a normal vector and a distance from the origin. The plane equation is Ax+By+Cz=D.    `camFrame.vz` is the view direction which I am using as the normal to the plane.  That provides the A,B,C.  And the plane is positioned at (`axis.p` dot `camFrame.vz`) units away from the eye.  That's D.  `axis.p` is the world space center of the gizmo.

When you "dot" this plane vector with a point (x,y,z,1), you get a number that is positive or negative if the point is on either side of the plane.  That's the equivalent to the value of Ax+By+Cz-D.  if it's zero, the point is on the plane.

renderlib/gesture/gesture.cpp
```diff
@@ -628,8 +629,15 @@ Gesture::drawCircle(glm::vec3 center,
     glm::vec3 p0 = center + xaxis * cosf(theta0) + yaxis * sinf(theta0);
     glm::vec3 p1 = center + xaxis * cosf(theta1) + yaxis * sinf(theta1);
 
-    graphics.addLine(Gesture::Graphics::VertsCode(p0, color, opacity, code),
-                     Gesture::Graphics::VertsCode(p1, color, opacity, code));
+    if (clipPlane) {
+      if (glm::dot(*clipPlane, glm::vec4(p0, 1.0)) > 0 && glm::dot(*clipPlane, glm::vec4(p1, 1.0)) > 0) {
+        graphics.addLine(Gesture::Graphics::VertsCode(p0, color, opacity, code),
+                         Gesture::Graphics::VertsCode(p1, color, opacity, code));
+      }
```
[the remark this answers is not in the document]
toloudis: Yeah this is admittedly crude but it works if the circles use small enough segments.   I'll refine it as needed for future usage.  Currently there's nothing else that needs to clip, yet.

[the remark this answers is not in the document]
toloudis: It's also possible to use gpu clip planes and have this all done in the shader with good interpolation, but it would result in potentially splitting up the geometry into more draw calls. Currently the code can accumulate lots of line segments into a single draw call even if they are drawing many different things.   And while things are still super cheap to draw right now, I didn't feel like routing all the data into the shader and splitting draw calls if they need different clip planes.

renderlib/RotateTool.cpp
ShrimpCryptid: > ...And the plane is positioned at (`axis.p` dot `camFrame.vz`) units away from the eye...

Clarification question, should this be units away from the origin?

toloudis: Yes, my mistake- it's in world space units from the origin, not from the eye.

ShrimpCryptid:
One more question, just as a check for a possible bug, if you're looking straight down an axis, will the ring that faces that axis be visible or clipped? Could be worth adding a slight offset to the plane just in case.

toloudis:
> One more question, just as a check for a possible bug, if you're looking straight down an axis, will the ring that faces that axis be visible or clipped? Could be worth adding a slight offset to the plane just in case.

It is very possible that it's clipped but in that case there is a white ring around the outside which would rotate about the same axis so there is no loss of functionality.  I would consider changing the test to be >= instead of > to account for when the ring is in the plane.

toloudis:
oops I merged in the wrong order and now this PR has extra changes in it

renderlib/CCamera.cpp
```diff
@@ -126,6 +126,7 @@ cameraManipulationDolly(const glm::vec2 viewportSize,
     camera.m_From += motion;
     camera.m_Target += targetMotion;
     camera.m_OrthoScale *= factor;
+    camera.Update();
```
toloudis, line 129: ignore; approved in previous PR and merged by accident into this one

renderlib/gesture/gesture.cpp
```diff
@@ -248,12 +248,18 @@ class ScopedGlVertexBuffer
...
   // Gesture draw spans across the entire window and it is not restricted to a single
   // viewport.
   if (this->verts.empty()) {
     clearCommands();
+
+    // TODO: do this clear only once if verts empty on consecutive frames?
+    // it would save some computation but this is really not a bottleneck here.
+    if (selection) {
+      selection->clear();
+    }
```
toloudis, line 262: ignore; approved in previous PR and merged by accident into this one

renderlib/io/FileReaderTIFF.cpp
```diff
@@ -139,7 +139,7 @@ readTiffNumScenes(TIFF* tiff, const std::string& filepath)
     LOG_WARNING << "Failed to read imagedescription of TIFF: '" << filepath << "';  interpreting as single channel.";
   }
 
-  std::string simagedescription = trim(imagedescription);
+  std::string simagedescription = trim(imagedescription ? imagedescription : "");
```
toloudis, line 142: ignore; approved in previous PR and merged by accident into this one

frasercl (approved):
Looks good! Your answers to @ShrimpCryptid's clarifying questions were very helpful to my understanding as well.

PR #172, opened 2024-03-23
Fix: clear selection buffer, and zooming in ortho mode

toloudis:
Est. time to review: <10 min

Three small bugfixes:
1. in some situations the rotation gizmo was still leaving data in the selection buffer even after it was disabled.  This fix ensures that the selection buffer is cleared more aggressively.
2. When zooming in orthographic camera mode, and then using the rotation gizmo, the view would become corrupted by showing the volume at the wrong scale.  Fixed by calling camera.Update() after  zoom/dolly.
3. Some tiff files have no imagedescription at all.  Agave was crashing on those files and this change allows them to be loaded.

Also dependabot put in some minor stuff into package-lock.json which is basically irrelevant to this PR (and that code is not being run or tested yet)

### DIFF
--- a/renderlib/gesture/gesture.h
+++ b/renderlib/gesture/gesture.h
@@ -397,7 +397,7 @@ struct Gesture
 
     // Gesture draw, called once per window update (frame) when the GUI draw commands
     // had been described in full.
-    void draw(struct SceneView& sceneView, const struct SelectionBuffer* selection);
+    void draw(struct SceneView& sceneView, struct SelectionBuffer* selection);
 
     // Pick a GUI element using the cursor position in Input.
     // Return a valid GUI selection code, SelectionBuffer::k_noSelectionCode
--- a/renderlib/CCamera.cpp
+++ b/renderlib/CCamera.cpp
@@ -126,6 +126,7 @@ cameraManipulationDolly(const glm::vec2 viewportSize,
     camera.m_From += motion;
     camera.m_Target += targetMotion;
     camera.m_OrthoScale *= factor;
+    camera.Update();
 
     // Consume gesture on button release
     Gesture::Input::reset(button);
--- a/renderlib/ViewerWindow.h
+++ b/renderlib/ViewerWindow.h
@@ -35,8 +35,6 @@ public:
     m_activeTool = (tool ? tool : &m_defaultTool);
 
     // clear out the buffer once.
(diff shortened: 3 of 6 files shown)


renderlib/gesture/gesture.h
```diff
@@ -397,7 +397,7 @@ struct Gesture
 
     // Gesture draw, called once per window update (frame) when the GUI draw commands
     // had been described in full.
-    void draw(struct SceneView& sceneView, const struct SelectionBuffer* selection);
+    void draw(struct SceneView& sceneView, struct SelectionBuffer* selection);
```
meganrm, line 400: what does this change do?

meganrm:
> Est. time to review: <10 min
> 
> Three small bugfixes:
> 
> 1. in some situations the rotation gizmo was still leaving data in the selection buffer even after it was disabled.  This fix ensures that the selection buffer is cleared more aggressively.
> 2. When zooming in orthographic camera mode, and then using the rotation gizmo, the view would become corrupted by showing the volume at the wrong scale.  Fixed by calling camera.Update() after  zoom/dolly.
> 3. Some tiff files have no imagedescription at all.  Agave was crashing on those files and this change allows them to be loaded.
> 
> Also dependabot put in some minor stuff into package-lock.json which is basically irrelevant to this PR (and that code is not being run or tested yet)

Given this isn't a language I'm familiar with:
#2 and 3 here seem really clear and I understand the fixes. 
the fix for #1 seems small too but I don't understand why those changes fixed the problem

renderlib/gesture/gesture.h
toloudis: The `clear()` method on SelectionBuffer actually modifies some internal data so it can't be considered const.  If a function calls any non-const methods on an object then the object can't be passed in as const.

toloudis:
>  the fix for problem 1 seems small too but I don't understand why those changes fixed the problem

I didn't completely prove this but I think the clear() was being called when there is no current opengl context and so issuing GL commands to clear a gpu buffer would be either a no op or undefined behavior.  Rather than call `makeCurrent/doneCurrent` to grab the opengl context, I opt to clear the buffer as part of the regular redraw cycle where we are doing other gl stuff.

blairlyons (approved):

meganrm (approved):

PR #184, opened 2024-05-02
Feature/flip volume

toloudis:
Estimated time to review:  ~15 min ?  A lot of code changes but all small.

Allows users to flip (mirror / invert) each of the three x,y,z axes for the volume.

AGAVE 1 6 0 2024-05-02 11-27-13

The internal storage uses -1 for a flipped axis, and 1 for an unflipped one.

The volume sampling shader has to switch its sampling wrap mode to "REPEAT" to make this work.

When adding a new feature like this, we need to make sure the feature is implemented in json save files (serialization), python script output, as well as client/server communication (aka Commands). 

In order to make it easier to add new serialized variables, without having to bump versions, I had to upgrade the json library I am using.   Disregard changes to json.hpp as it is a third party file.

### DIFF
--- a/agave_app/AppearanceSettingsWidget.cpp
+++ b/agave_app/AppearanceSettingsWidget.cpp
@@ -165,7 +165,6 @@ QAppearanceSettingsWidget::QAppearanceSettingsWidget(QWidget* pParent,
   m_zscaleSpinner->setToolTip(tr("Scale volume in Z dimension"));
   m_zscaleSpinner->setDecimals(6);
   m_zscaleSpinner->setValue(1.0);
-  m_zscaleSpinner->setMinimum(-m_zscaleSpinner->maximum());
   scaleSectionLayout->addWidget(m_zscaleSpinner, 2, 1);
   m_zFlipCheckBox = new QCheckBox("Flip");
   m_zFlipCheckBox->setStatusTip(tr("Invert volume in Z dimension"));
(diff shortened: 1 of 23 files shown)


renderlib/json/json.hpp
toloudis, line 1: can disregard diff in this file as it is a third party include which I updated

agave_app/Serialize.h
```diff
@@ -106,28 +107,29 @@ struct ViewerState
...
-                                 camera,
-                                 backgroundColor,
-                                 boundingBoxColor,
-                                 showBoundingBox,
-                                 channels,
-                                 density,
-                                 lights,
-                                 capture,
-                                 showScaleBar)
+  NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(ViewerState,
```
toloudis, line 115: comment why this macro is here and why it usees WITH_DEFAULT

renderlib/ImageXYZC.cpp
```diff
@@ -79,9 +80,21 @@ ImageXYZC::maxPixelDimension() const
...
-  m_scaleX = x;
-  m_scaleY = y;
-  m_scaleZ = z;
+  m_scaleX = abs(x);
+  m_scaleY = abs(y);
+  m_scaleZ = abs(z);
+}
+
+void
+ImageXYZC::setVolumeAxesFlipped(float x, float y, float z)
```
toloudis, line 89: this really should be storing ints or booleans

agave_app/AppearanceSettingsWidget.cpp
```diff
@@ -369,6 +384,34 @@ QAppearanceSettingsWidget::createSkyLightingControls()
   return section;
 }
 
+void
+QAppearanceSettingsWidget::OnFlipX(bool value)
```
meganrm, line 388: just making a note of my suggestion to change these 3 functions to one that takes in an `axis` value to reduce the copied code

meganrm (approved):
Some small suggestions came up in our review meeting, but overall seems good

rugeli (approved):
Looks great!

PR #188, opened 2024-05-07
draw bounding box in raymarch render mode!

toloudis:
Estimated review time:  ~10-15 min.

For some reason, in ray-march render mode, bounding boxes were never showing up.  This is because it was never implemented.
This code adds bounding box drawing in raymarch (non-pathtraced) mode.  This render mode is very much the same as what our web volumeviewer displays.

### DIFF
--- a/renderlib/graphics/RenderGL.cpp
+++ b/renderlib/graphics/RenderGL.cpp
@@ -6,6 +6,7 @@
 #include "Logging.h"
 #include "RenderSettings.h"
 #include "gl/Image3D.h"
+#include "gl/Util.h"
 
 #include <iostream>
 
@@ -17,6 +18,7 @@ RenderGL::RenderGL(RenderSettings* rs)
   , m_h(0)
   , m_renderSettings(rs)
   , m_scene(nullptr)
+  , m_boundingBoxDrawable(nullptr)
   , m_status(new CStatus)
 {
   mStartTime = std::chrono::high_resolution_clock::now();
@@ -40,6 +42,7 @@ RenderGL::initialize(uint32_t w, uint32_t h)
   glEnable(GL_BLEND);
   glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
 
+  m_boundingBoxDrawable = new BoundingBoxDrawable();
   if (m_scene && m_scene->m_volume) {
     initFromScene();
   }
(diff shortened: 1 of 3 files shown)


renderlib/graphics/RenderGL.cpp
```diff
@@ -100,11 +103,60 @@ RenderGL::renderTo(const CCamera& camera, GLFramebufferObject* fbo)
...
   doClear();
   if (haveScene) {
-    m_image3d->render(camera, m_scene, m_renderSettings);
+    drawSceneObjects(camera);
   }
   fbo->release();
 }
 
+void
+RenderGL::drawSceneObjects(const CCamera& camera)
```
toloudis, line 112: most of the lines of code in here are just getting the bounding box's transform matrix to position and scale it correctly in the world with the volume being drawn.  This is repeated code from the pathtrace renderer and can be DRY refactored at some point.

renderlib/graphics/RenderGL.cpp
```diff
@@ -100,11 +103,60 @@ RenderGL::renderTo(const CCamera& camera, GLFramebufferObject* fbo)
...
+  // it helps to imagine these transforming the space in reverse order
+  // (first translate by 1.0, and then scale down)
+  glm::mat4 bboxModelMatrix = glm::scale(glm::mat4(1.0f), scales);
+  bboxModelMatrix = glm::translate(bboxModelMatrix, glm::vec3(1.0, 1.0, 1.0));
+  glm::mat4 viewMatrix(1.0);
+  glm::mat4 projMatrix(1.0);
+  camera.getProjMatrix(projMatrix);
+  camera.getViewMatrix(viewMatrix);
+
+  if (m_scene->m_material.m_showBoundingBox) {
```
ShrimpCryptid, line 143: I'm seeing some other flags get enabled in RenderGLPT.cpp that aren't here (`GL_DEPTH_TEST`, Depth Mask, Color Mask, etc.). Are those not required here because we can directly draw the lines in the scene, and we don't have to worry about sorting with the pathtraced volumes?

renderlib/graphics/RenderGL.cpp
```diff
@@ -100,11 +103,60 @@ RenderGL::renderTo(const CCamera& camera, GLFramebufferObject* fbo)
...
+                        m_scene->m_material.m_boundingBoxColor[2],
+                        1.0);
+    m_boundingBoxDrawable->drawLines(projMatrix * viewMatrix * bboxModelMatrix, bboxColor);
+    if (m_scene->m_showScaleBar && camera.m_Projection != ProjectionMode::ORTHOGRAPHIC) {
+      m_boundingBoxDrawable->updateTickMarks(scales, maxPhysicalDim);
+      m_boundingBoxDrawable->drawTickMarks(projMatrix * viewMatrix * bboxModelMatrix, bboxColor);
+    }
+  }
+
+  m_image3d->render(camera, m_scene, m_renderSettings);
```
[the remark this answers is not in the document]
toloudis: just to separate the "draw stuff" from all the extra framebuffer setup and teardown that wraps it.   Also a bit of DRY since it does appear in two separate functions.

renderlib/graphics/RenderGL.cpp
toloudis: Yes, those settings are differences in how the pathtrace renderer draws itself.  Raymarch is simpler because it's drawing a box geometry, whereas pathtrace is basically drawing a flat fullscreen buffer that needs to be composited correctly as if it is inside the bounding box lines.

ShrimpCryptid (approved):

interim17 (approved):
I'm definitely out of my depth here! Don't have much to add at this point, but it's a great stretch for me to read through this code.

PR #190, opened 2024-05-11
Installers and load via agave:// url protocol

toloudis:
Estimated time to review: medium

We wanted to support integration with BioFileFinder, such that it can have an "Open With AGAVE" command.  

In order to do this, we need to set up url protocol handlers for the custom `agave://` protocol, so that the BFF web app can just "navigate" to an agave link, and then agave will open automatically.  Protocol handlers are set up differently for mac and windows.

In order to have the url handlers registered on the user's system, we make that happen via installers for MacOS and Windows.  **YES, AGAVE FINALLY HAS AN INSTALLER FOR MACOS x86-64, ARM64, AND WINDOWS.**

The other code changes are around receiving the volume file in agave via url, and parsing the volume file path from the url so agave can open it.

Part of this change also adds a simple command line arg:  `agave --load url_or_localpath`.
Urls must be of the form: `agave://?url=url_or_localpath`

Note for MacOS: our code is self-signed and not notarized, and as a result the MacOS installer for ARM64 (Apple silicon) results in an apparently damaged app. It's actually just not passing the Apple quarantine, and for now you have to run an `xattr` command to tell macos that the app is safe.

Much credit to @saeliddp for a bunch of work on the macos installer portions of this PR, and testing.


CMake/Info.plist.cmake
```diff
@@ -22,6 +22,17 @@
 	<string>@PROJECT_VERSION@</string>
 	<key>CFBundleShortVersionString</key>
 	<string>@PROJECT_VERSION@</string>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+		<key>CFBundleURLName</key>
+		<string>com.aics.agave</string>
```
toloudis, line 29: maybe `org.allencell.agave`

CMakeLists.txt
```diff
@@ -140,67 +141,60 @@ CONFIGURE_FILE(${CMAKE_SOURCE_DIR}/CMake/conf.py.cmake ${CMAKE_SOURCE_DIR}/docs/
 # #####################
 find_package(Qt6QTiffPlugin 6.5.3 REQUIRED PATHS ${Qt6Gui_DIR})
 
-# copy all dlls into place.
+set(CPACK_PACKAGE_VENDOR "Allen Institute for Cell Science")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "AGAVE is a viewer for multi-dimensional microscopy images.")
```
toloudis, line 145:
```suggestion
set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "AGAVE is a viewer for 4D multichannel microscopy images, using physically based lighting and rendering.")

```

CMakeLists.txt
```diff
@@ -140,67 +141,60 @@ CONFIGURE_FILE(${CMAKE_SOURCE_DIR}/CMake/conf.py.cmake ${CMAKE_SOURCE_DIR}/docs/
...
     DESTINATION ${INSTALLDIR}
     CONFIGURATIONS Debug
   )
+
   install(FILES
     ${PROJECT_SOURCE_DIR}/renderlib/assets/fonts/Arial.ttf
     ${PROJECT_SOURCE_DIR}/renderlib/assets/fonts/Courier_New.ttf
     DESTINATION ${INSTALLDIR}/assets/fonts/
   )
+  #install(DIRECTORY ${PROJECT_SOURCE_DIR}/agave_app/icons/ DESTINATION ${INSTALLDIR}/icons/)
```
toloudis, line 196:
```suggestion
  

```

saeliddp (approved):

saeliddp:
Where are we planning to document that the macOS ARM installer requires the extra `sudo xattr -rd com.apple.quarantine /Applications/agave.app` step?

CMakeLists.txt
```diff
@@ -222,6 +214,29 @@ if(WIN32)
     DESTINATION ${INSTALLDIR}/styles/
   )
 
+  set(CPACK_GENERATOR "NSIS")
+  set(CPACK_PRE_BUILD_SCRIPTS "${CMAKE_SOURCE_DIR}/CMake/WindowsPreBuild.cmake")
+  set(CPACK_PACKAGE_INSTALL_DIRECTORY "AGAVE ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
```
ShrimpCryptid, line 219: Should all the AGAVE version install to different folders, or should they all share the same on?

If someone has an existing installation of AGAVE, is the installer smart enough to delete the old installation? Will there be conflicts in the `agave://` url protocol registration if a new version of AGAVE is installed?

toloudis:
> Where are we planning to document that the macOS ARM installer requires the extra `sudo xattr -rd com.apple.quarantine /Applications/agave.app` step?

I definitely intend to document it in release notes, on the main agave web page at allencell.org, and maybe in the agave docs as a separate PR.

CMakeLists.txt
toloudis: currently this will always overwrite the url protocol handler with the most recently installed agave.

toloudis: On windows, we don't automatically uninstall previous agave. It's true that they could mess up the protocol handler by installing a new version and then uninstalling the old.  I'll think about how best to handle that.  At least we can warn them to uninstall FIRST before installing.

toloudis: Maybe I could just make the reg key have the version number in it.  will have to test

aswallace (approved):
Tested .dmg installation with M3 chip via (what I assume to be) the latest GH actions artifacts, confirmed I'm able to open an S3-stored zarr file via BFF.

Also attempted installing from source;  needed to `brew install nasm` before I could run `cmake ..`, noted that it's mentioned in the Windows instructions that I skimmed over bc Mac. Maybe worth adding in Mac `brew` part of readme as well?

PR #193, opened 2024-06-08
Add option to interpolate volume sampling

### LINKED ISSUE #182: interpolation on/off
## Use Case

linear interpolation of volume samples can lead to visual artifacts that are not desirable especially with segmentations

## Solution

checkbox to switch interpolation on/off.  unfortunately this can not be per channel due to the packing of multi channels into single texture.

toloudis:
Review size estimate: 
medium. - many files are touched but it's not a whole lot of code.  Most of it is just boilerplate for adding this new feature to all the parts of the app.

Description

Resolves #182 

The pathtracer's paths may or may not terminate at the centers of voxels.  By default, we sample the volume using linear filtering so that the sample is a weighted sum of neighboring voxels.

This code change allows an option to switch off the linear sampling and use nearest-neighbor sampling instead.  Basically this makes the rendering look blockier - if you zoom in you can easily see edges of voxels.  This "nearest" lookup can be an improvement for segmentations in which you want to see those actual voxel edges, or even for lower resoluton raw image data.  Unfortunately it has to be a global setting that is applied to all channels because of how the channels are interleaved in GPU memory.

The changes cover:
* adding a checkbox to the GUI
* saving to json, and loading from json, and initializing the GUI when loading
* adding the setting to the Python API, saving to python script, and adding a "Command" that contains the implementation.  The command is similarly added to the javascript API.

The changes are pretty well split up between the three separate commits, if interested.

### DIFF
--- a/agave_app/python/RenderInterface.h
+++ b/agave_app/python/RenderInterface.h
@@ -70,4 +70,9 @@ public:
   virtual int BackgroundColor(float, float, float) = 0;
   virtual int SetIsovalueThreshold(int32_t, float, float) = 0;
   virtual int SetControlPoints(int32_t, std::vector<float>) = 0;
+  virtual int SetBoundingBoxColor(float, float, float) = 0;
+  virtual int ShowBoundingBox(int32_t) = 0;
+  virtual int ShowScaleBar(int32_t) = 0;
+  virtual int SetFlipAxis(int32_t, int32_t, int32_t) = 0;
+  virtual int SetInterpolation(int32_t) = 0;
 };
--- a/agave_app/python/pyrenderer.h
+++ b/agave_app/python/pyrenderer.h
@@ -104,6 +104,8 @@ public:
   virtual int SetBoundingBoxColor(float, float, float);
   virtual int ShowBoundingBox(int32_t);
   virtual int ShowScaleBar(int32_t);
+  virtual int SetFlipAxis(int32_t, int32_t, int32_t);
+  virtual int SetInterpolation(int32_t);
 
 protected:
   void init();
--- a/agave_pyclient/agave_pyclient/agave.py
+++ b/agave_pyclient/agave_pyclient/agave.py
@@ -937,6 +937,18 @@ class AgaveRenderer:
         # 46
         self.cb.add_command("SET_FLIP_AXIS", x, y, z)
 
+    def set_interpolation(self, x: int):
+        """
+        Set the volume sampling interpolation mode
+
+        Parameters
+        ----------
+        x: int
+            0 for nearest, 1 for linear
+        """
+        # 47
+        self.cb.add_command("SET_INTERPOLATION", x)
+
     def batch_render_turntable(
         self, number_of_frames=90, direction=1, output_name="frame", first_frame=0
     ):
--- a/renderlib/graphics/ImageXyzcGpu.cpp
+++ b/renderlib/graphics/ImageXyzcGpu.cpp
@@ -199,3 +199,12 @@ ImageGpu::updateLutGpu(int channel, ImageXYZC* img)
 {
   m_channels[channel].updateLutGpu(channel, img);
 }
+
+void
+ImageGpu::setVolumeTextureFiltering(bool linear)
+{
+  glBindTexture(GL_TEXTURE_3D, m_VolumeGLTexture);
+  glTexParameteri(GL_TEXTURE_3D, GL_TEXTURE_MIN_FILTER, linear ? GL_LINEAR : GL_NEAREST);
+  glTexParameteri(GL_TEXTURE_3D, GL_TEXTURE_MAG_FILTER, linear ? GL_LINEAR : GL_NEAREST);
+  glBindTexture(GL_TEXTURE_3D, 0);
+}
--- a/renderlib/graphics/gl/Image3D.cpp
+++ b/renderlib/graphics/gl/Image3D.cpp
@@ -224,7 +224,7 @@ Image3D::setSize(const glm::vec2& xlim, const glm::vec2& ylim)
 }
 
 void
-Image3D::prepareTexture(Scene& s)
+Image3D::prepareTexture(Scene& s, bool useLinearInterpolation)
 {
   auto startTime = std::chrono::high_resolution_clock::now();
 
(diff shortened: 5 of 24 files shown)


agave_app/python/RenderInterface.h
```diff
@@ -70,4 +70,9 @@ class RenderInterface
   virtual int BackgroundColor(float, float, float) = 0;
   virtual int SetIsovalueThreshold(int32_t, float, float) = 0;
   virtual int SetControlPoints(int32_t, std::vector<float>) = 0;
+  virtual int SetBoundingBoxColor(float, float, float) = 0;
+  virtual int ShowBoundingBox(int32_t) = 0;
+  virtual int ShowScaleBar(int32_t) = 0;
+  virtual int SetFlipAxis(int32_t, int32_t, int32_t) = 0;
+  virtual int SetInterpolation(int32_t) = 0;
```
toloudis, line 77: the above functions were missing, so I just threw them in for consistency. (This class ideally is the complete command api) This code in here is actually dead code currently and probably should be removed in a separate PR.  It was meant as a way to do a cleaner python integration that didn't require you to run agave separately as a server process, but never got it fully working.

renderlib/graphics/ImageXyzcGpu.cpp
```diff
@@ -199,3 +199,12 @@ ImageGpu::updateLutGpu(int channel, ImageXYZC* img)
...
   m_channels[channel].updateLutGpu(channel, img);
 }
+
+void
+ImageGpu::setVolumeTextureFiltering(bool linear)
+{
+  glBindTexture(GL_TEXTURE_3D, m_VolumeGLTexture);
+  glTexParameteri(GL_TEXTURE_3D, GL_TEXTURE_MIN_FILTER, linear ? GL_LINEAR : GL_NEAREST);
+  glTexParameteri(GL_TEXTURE_3D, GL_TEXTURE_MAG_FILTER, linear ? GL_LINEAR : GL_NEAREST);
+  glBindTexture(GL_TEXTURE_3D, 0);
```
toloudis, line 209: the above is the core implementation of the feature for the pathtracer -  it tells the gpu to sample the volume differently based on the setting

renderlib/graphics/gl/Image3D.cpp
```diff
@@ -253,9 +253,9 @@ Image3D::prepareTexture(Scene& s)
   // glGenTextures(1, &_textureid);
   glBindTexture(GL_TEXTURE_3D, m_textureid);
   check_gl("Bind texture");
-  glTexParameteri(GL_TEXTURE_3D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+  glTexParameteri(GL_TEXTURE_3D, GL_TEXTURE_MIN_FILTER, useLinearInterpolation ? GL_LINEAR : GL_NEAREST);
   check_gl("Set texture min filter");
-  glTexParameteri(GL_TEXTURE_3D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
+  glTexParameteri(GL_TEXTURE_3D, GL_TEXTURE_MAG_FILTER, useLinearInterpolation ? GL_LINEAR : GL_NEAREST);
```
toloudis, line 258: these changes are the implementation for the "ray march" renderer, same as what the path trace renderer has to do

toloudis:
Taking this back to draft to fix the build... gh-actions build is having trouble with this

renderlib/io/CMakeLists.txt
```diff
@@ -28,22 +28,12 @@ add_subdirectory(libCZI)
...
-FetchContent_MakeAvailable(zstd)
-
 # tensorstore dependency for renderlib
 set(TENSORSTORE_USE_SYSTEM_TIFF ON)
 FetchContent_Declare(
 	tensorstore
-	URL "https://github.com/google/tensorstore/archive/refs/tags/v0.1.58.tar.gz"
-	URL_HASH SHA256=e916c9dc0ca9a7f272fc975b76ea06c452f921e65a8b201fc45b4ea96261a6ad
+	GIT_REPOSITORY https://github.com/google/tensorstore.git
+	GIT_TAG v0.1.61
```
toloudis, line 36: The changes in this file are a throw-in to fix builds on github-actions, separate from the intent of the PR.

agave_app/python/pyrenderer.h
```diff
@@ -104,6 +104,8 @@ class OffscreenRenderer : public RenderInterface
   virtual int SetBoundingBoxColor(float, float, float);
   virtual int ShowBoundingBox(int32_t);
   virtual int ShowScaleBar(int32_t);
+  virtual int SetFlipAxis(int32_t, int32_t, int32_t);
```
meganrm, line 107: is this another adding something that should have been there? I thought the flip axis was an feature you already added

agave_pyclient/agave_pyclient/agave.py
```diff
@@ -937,6 +937,18 @@ def set_flip_axis(self, x: int, y: int, z: int):
...
         self.cb.add_command("SET_FLIP_AXIS", x, y, z)
 
+    def set_interpolation(self, x: int):
+        """
+        Set the volume sampling interpolation mode
+
+        Parameters
+        ----------
+        x: int
+            0 for nearest, 1 for linear
```
meganrm, line 947: just curious: is it just always easier to deal with ints? if I was writing this in typescript I would use the names "nearest" and "linear" so that it was easy to know what the settings were throughout the code. Or have it be a bool, since this feature is just enabling or disabling a method

meganrm (approved):
seems good, I did just have one question about your parameter choice, but not blocking

agave_pyclient/agave_pyclient/agave.py
toloudis: Yes from an API design standpoint this is not good. 

The reason it's like this is because these commands that come from the python client or the js web client are getting converted to binary data stream to be sent to the c++ code via network (websocket).  And the client/server system I set up can deal with basic types like ints, floats, and arrays of ints and floats.  (Strings are also done as array of bytes)

So when you see these ints-as-booleans creep into the apis, it's just because it keeps things consistent across python, c++, and js codebases with the identical function signature.
A nice-to-have code change would be to set up good symbolic enums for the functions that take data types like this one, or booleans, and convert internally to the integer types. Then the external python and js apis could be more user friendly.

agave_app/python/pyrenderer.h
toloudis: yes, this was just added for completeness because I noticed it was missing.  Shouldn't even affect any runtime behaviors.

ascibisz (approved):
I only kinda understand the nearest neighbor vs linear interpolation big picture stuff, but as far as the code doing what you say it should be doing, it looks good to me!

toloudis:
> I only kinda understand the nearest neighbor vs linear interpolation big picture stuff, but as far as the code doing what you say it should be doing, it looks good to me!

For segmentations, which have binary information 0 or 1 everywhere, interpolation can make the edges look fuzzier and less well defined.  It can also have weird visual halo artifacts due to not knowing exactly where the surface is or what direction it's pointing in.
For fuzzier raw data that is more continuous and noisy in nature, the interpolation can look totally fine.
Here's a segmentation image that shows the difference:
With interpolation: (the old way)
testinterp
Without interpolation:
testnointerp

PR #195, opened 2024-06-12
update tensorstore build

toloudis:
bump tensorstore version and remove explicit zstd, to fix build issues with github-actions

### DIFF
--- a/renderlib/io/CMakeLists.txt
+++ b/renderlib/io/CMakeLists.txt
@@ -32,7 +32,7 @@ add_subdirectory(libCZI)
 set(TENSORSTORE_USE_SYSTEM_TIFF ON)
 FetchContent_Declare(
 	tensorstore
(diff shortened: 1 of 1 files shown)


rugeli (approved):
curious: by removing the explicit `zstd`, I'm assuming the newer version of TensorStore manages `zstd` dependency on its own? Just wondering if this removal could potentially lead to any conflicts?

Otherwise lgtm!

toloudis:
> curious: by removing the explicit `zstd`, I'm assuming the newer version of TensorStore manages `zstd` dependency on its own? Just wondering if this removal could potentially lead to any conflicts?
> 
> Otherwise lgtm!

Yes, tensorstore can figure out how to pull in zstd also. However, I actually have instructions in the README and part of the github actions that run automated builds, that also brings in zstd through other ways.  I think the FetchContent here was redundant and unnecessary.

meganrm (approved):

PR #197, opened 2024-06-12
Axis helper indicator in viewport

toloudis:
Add a XYZ axis indicator at lower left corner of viewport.
X is red, Y is green, Z is blue.
Axes will follow camera rotation angle, and use same projection, perspective or orthographic.
The indicator can be turned on and off by checkbox.  

Currently there is no saved state for this, as it is intended only for interactive viewport use.
It will show up in renders only if enabled in the viewport. 
It is not a part of the remote command API and therefore can not be turned on/off via Python script.

The control is enabled/disabled by the rightmost toolbar button that shows an icon of the X/Y/Z triple axis.
Perspective:
AGAVE 1 6 4 2024-07-16 13-58-51
Orthographic:
AGAVE 1 6 4 2024-07-16 13-59-17


agave_app/ViewToolbar.cpp
```diff
@@ -101,6 +101,16 @@ ViewToolbar::ViewToolbar(QWidget* parent)
   rightViewButton->setFocusPolicy(Qt::NoFocus);
   toolbarLayout->addWidget(rightViewButton);
 
+  toolbarLayout->addItem(new QSpacerItem(spacing, 0, QSizePolicy::Fixed, QSizePolicy::Expanding));
```
rugeli, line 104: It took me some time to learn about Qt Widgets and seems they are well used here!

rugeli (approved):
Thank you for the comments, they are useful for me to understand the code! 
 
 Regarding the note `It is not a part of the remote command API and therefore can not be turned on/off via Python script.` is this the intended functionality or do we have plans to integrate this feature into the API in the future? 
 
 Overall looks well structured and modular to me!

agave_app/ViewToolbar.cpp
toloudis: Yes, the whole user interface is built with them!

toloudis:
> Regarding the note `It is not a part of the remote command API and therefore can not be turned on/off via Python script.` is this the intended functionality or do we have plans to integrate this feature into the API in the future?
> 

This is something I'm not sure is needed by users but would be easy enough to add later. I also felt that adding it now would have made this PR a lot bigger and is probably easier to review if I do it separately.

renderlib/AxisHelperTool.cpp
```diff
@@ -0,0 +1,128 @@
...
+  // translate this in screen space toward the lower left corner.
+  // note that this is still drawn in perspective and will have fov distortion.
+  // Ideally we want to give this its own viewport as if it were centered.
+  // (Or translate the center of projection!)
+  static constexpr float XPOS_NDC = -0.45f;
+  static constexpr float YPOS_NDC = -0.45f;
+
+  // Draw the manipulator to be at some constant size on screen
+  float scaleFactor = 0.333; // make it smaller
+  float scale = scaleFactor;
```
ascibisz, line 52: why do we set `scale` to `scaleFactor` here?

toloudis: This is just to initialize it to something.  The if/else below is going to reset it but I wanted to give it some valid init value here.

PR #201, opened 2024-06-26
Refactor: move opengl code out of the gesture module

toloudis:
As a precursor to potentially replacing OpenGL with a different graphics API like vulkan, wgpu or OptiX, I need to isolate the OpenGL code away from the reusable generic graphics code.

This PR has no functionality changes at all and simply separates the OpenGL code that was part of the `gesture` library.  (The gesture library is used for in-viewport mouse manipulation and also for drawing certain widgets like axis indicators, rotation tools, scale bars)

Key changes:
* the Font module now is split into two separate modules:  cpu-side font bitmap generation, and opengl texture creation.
* the Gesture.Graphics library had a `draw` and `pick` function that have opengl implementations, which have all been moved into GestureRendererGL. The idea is that the Gesture Renderer knows how to use OpenGL to render a "scene" composed of Gesture::Graphics Commands.
* the Gesture draw commands were using opengl enumerations when they should be using a api-independent enum.

The refactor is pretty naive and there is probably a reasonable followup that makes the refactor better.  (For example, the draw and pick functions are changing things in the gesture internals that should probably be done outside of draw and pick.)


renderlib/graphics/GestureGraphicsGL.cpp
toloudis, line 1: basically this file is composed of all the stuff that was moved out of gesture/gesture.cpp

meganrm (approved):
seems ok to me: if it's working. I don't have enough familiarity of the overall structure to suggest any additional organization

interim17 (approved):
Echo what Megan said! LGTM

PR #202, opened 2024-06-26
fix time slider updates for raymarch mode

### LINKED ISSUE #198: Time slider does not update view in Ray March Blending mode
## Description

- Changing the time slider does not cause the view to update in Ray march blending mode.

## Expected Behavior

_What did you expect to happen instead?_

## Reproduction

- Open https://dev-aics-dtp-001.int.allencell.org/dan-data/3500005818_20230811__20x_Timelapse-02(P27-E7).ome.zarr
- Switch to Ray March Blending mode
- Change the time slider to a different value.
- The viewport will only update if the view is changed back to Path trace mode.

## Environment

- Windows 10
- AGAVE v1.6.4

toloudis:
fix #198 

Time slider change was not causing updates in the raymarch rendered image.

When we do a time slider change, we completely replace the volume stored in the scene.

But the raymarch volume representation was keeping its own reference to a volume.  So it never knew that the scene's volume was different.  

This PR just has the Image3D always use a volume given to it from the Scene.

### DIFF
--- a/renderlib/graphics/RenderGL.cpp
+++ b/renderlib/graphics/RenderGL.cpp
@@ -215,8 +215,8 @@ RenderGL::initFromScene()
 {
   delete m_image3d;
 
-  m_image3d = new Image3D(m_scene->m_volume);
-  m_image3d->create();
+  m_image3d = new Image3D();
+  m_image3d->create(m_scene->m_volume);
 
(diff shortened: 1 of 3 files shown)


ShrimpCryptid (approved):
LGTM!

PR #203, opened 2024-07-08
Refactor for picking

toloudis:
Two small changes here.
1. Previous refactor #201 left behind some small bookkeeping that really should have been factored out.  This does that, in the pick function.  Pick now returns a boolean (as before) and also takes an in/out parameter of the selection code which is the actual pick result.
2. As seen here, https://github.com/allen-cell-animated/agave/pull/201/files#diff-c5c88e2609315d21a31aad6d4473e934233b5cdd1cc64d6c1a0996929836eb3cL306 , the previous refactor introduced a second k_noSelectionCode in a second place.  This change restores the value and makes sure it is only defined in one place.

### DIFF
--- a/renderlib/graphics/GestureGraphicsGL.cpp
+++ b/renderlib/graphics/GestureGraphicsGL.cpp
@@ -249,7 +249,7 @@ SelectionBuffer::clear()
   glBindFramebuffer(GL_FRAMEBUFFER, frameBuffer);
   {
     glViewport(0, 0, resolution.x, resolution.y);
-    uint32_t clearcode = SelectionBuffer::k_noSelectionCode;
+    uint32_t clearcode = Gesture::Graphics::k_noSelectionCode;
     glClearColor(((clearcode >> 0) & 0xFF) / 255.0,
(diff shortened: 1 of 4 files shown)


renderlib/graphics/GestureGraphicsGL.cpp
```diff
@@ -413,16 +413,8 @@ bool
...
+                        uint32_t& selectionCode)
 {
-  // If we are in mid-gesture, then we can continue to use the retained selection code.
-  // if we never had anything to draw into the pick buffer, then we didn't pick anything.
-  // This is a slight oversimplification because it checks for any single-button release
-  // or drag: two-button drag gestures are not handled well.
-  if (input.clickEnded() || input.isDragging()) {
-    return graphics.m_retainedSelectionCode != SelectionBuffer::k_noSelectionCode;
-  }
-
```
toloudis, line 425: this piece of logic is lifted out of here and put in ViewerWindow just before `pick` is called.

ShrimpCryptid (approved):

frasercl (approved):

PR #209, opened 2024-08-08
allow dark mode / light mode switching

toloudis:
Review time: medium/small
Most of the modified files are just the icons.

Description:
Set up dark mode / light mode, with a second set of icons.

Ensure all icon switching can happen via a app-level stylesheet.

In particular the dual-state button requires a special "custom Qt Property" which can then be referenced in the stylesheet.

macos screenshot example:

Dark mode:
AGAVE 1 6 4 2024-08-08 16-19-38

Light mode:
AGAVE 1 6 4 2024-08-08 16-19-59


agave_app/Controls.h
```diff
@@ -20,53 +20,48 @@
...
     if (state == 1) {
-      setIcon(icon2);
       setToolTip(tooltip2);
       setStatusTip(statustip2);
     } else {
-      setIcon(icon1);
       setToolTip(tooltip1);
       setStatusTip(statustip1);
     }
+    style()->polish(this);
```
toloudis, line 64: this line nudges the Qt style update code to respond to a property change so that it can refresh the icon

ShrimpCryptid: Super interesting that you're able to specify the icons by state in the stylesheet!

agave_app/agaveGui.cpp
```diff
@@ -37,12 +37,52 @@
 
 #include <filesystem>
 
+const QString darkStyleSheet = R"(
```
ShrimpCryptid, line 40: What does the `R` syntax here do?

agave_app/icons/dark/frameView.svg
ShrimpCryptid, line 1: Maybe out of scope of this PR, but I find this icon hard to parse because its scale is very different from the other UI icons. Happy to take a swing at editing this if you'd like.

ShrimpCryptid (approved):
Code changes look good! Left a few minor comments and one about the icon readability.

agave_app/icons/dark/linked.png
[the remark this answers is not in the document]
toloudis: no reason other than that I found this icon first and brought in the svgs later.  I'm up for changing it so they are all SVG at some point, that would be nice.

agave_app/agaveGui.cpp
toloudis: R lets you do a inline multiline string literal without having to escape any characters.

agave_app/icons/dark/frameView.svg
toloudis: yeah, feel free to thicken it or whatever.  The icons were sourced a bit randomly.  It would be good to have someone like Thao or UX in general do a pass and unify everything, validate whether the icons are doing a good job etc.  But I think they are not too controversial in general.

agave_app/agaveGui.cpp
```diff
@@ -1216,3 +1256,25 @@ agaveGui::appToViewerState()
...
+    if (m_colorScheme == colorScheme) {
+      return;
+    }
+    if (colorScheme == Qt::ColorScheme::Dark) {
+      setStyleSheet(darkStyleSheet);
+      LOG_DEBUG << "ThemeChange to Dark";
+    } else if (colorScheme == Qt::ColorScheme::Light) {
+      setStyleSheet(lightStyleSheet);
+      LOG_DEBUG << "ThemeChange to Light";
+    }
```
meganrm, line 1276: just curious: why is this code more or less a repeat of what's above? in lines 80-84. is there more than one way the theme gets changed?

agave_app/agaveGui.cpp
meganrm: is there a way this still gets linted to avoid typos?

meganrm (approved):
seems good, I have questions but they're mostly for my benefit

agave_app/agaveGui.cpp
toloudis: lines 80-84 are a one-time initialization.  This code here handles dynamic changes.  If I had to repeat the logic one more time then I'd refactor, or if the logic was any more complicated than this.  Maybe I'll change it so that it reads something like setStyleSheet(getStyleSheetString(colorScheme))

agave_app/agaveGui.cpp
toloudis: sadly, no.  There might be a future PR where the stylesheet gets much bigger and lives in its own resource file and then there will at least be a chance - but it's a special Qt flavor of CSS and probably wouldn't lint like real css.  Not sure if Qt provides any tools for this.

Issue #180, opened 2024-05-01
rendering gets out of sync


toloudis:
## Description

some renders appear to complete with too few samples

## Expected Behavior

all images get same number of samples

## Reproduction

set render resolution low like 256x256 and number of samples low like 32
then render a timelapse
some images will be undersampled.

## Environment

windows, gtx 4090

PR #210, opened 2024-08-13
Fix render thread synchronization problems

toloudis:
Review time: medium.  I am happy to have meeting time to explain this one for deeper understanding.

Addresses #180  : when rendering pathtrace iterations are fast, the main thread can not keep up with the pace of images being returned from the render thread.  This was causing renders to finish with wrong numbers of samples, and the progress bar could end up being wrong.

The fix is to count up progress toward completion inside of the render thread (instead of on the main thread).  Also the render thread now "emits" a special message when a rendered frame is complete, so the main thread can respond to that particular message and save the image and start the next one.

This is a "barebones" fix but paves the way for actual progress update messages to come out of the render thread also.

The render thread is basically contained in `renderer.h` and `renderer.cpp`.  The main thread controls the render thread from `RenderDialog.cpp`.


agave_app/renderer.h
```diff
@@ -30,6 +29,58 @@ class IRenderWindow;
...
+    outputDir = other.outputDir;
+    filenamePrefix = other.filenamePrefix;
+    width = other.width;
+    height = other.height;
+    renderDuration = other.renderDuration;
+    startTime = other.startTime;
+    endTime = other.endTime;
+  }
+};
+
```
toloudis, line 83: all the above changes were mainly to move these data structures into this class when they were previously in RenderDialog.

agave_app/renderer.h
```diff
@@ -30,6 +29,58 @@ class IRenderWindow;
...
 class Scene;
 
+// serialized so permanent?
+enum eRenderDurationType
+{
+  TIME = 0,
+  SAMPLES = 1
+};
+
+struct RenderDuration
```
toloudis, line 39: RenderDuration is a newly introduced data structure to encapsulate the data needed to determine if a render is finished or not.

agave_app/renderer.h
```diff
@@ -47,7 +98,8 @@ class Renderer
                  const LoadSpec& loadSpec,
                  // rendererMode ignored if renderer is non-null
                  renderlib::RendererType rendererMode = renderlib::RendererType_Pathtrace,
-                 QOpenGLContext* glContext = nullptr);
+                 QOpenGLContext* glContext = nullptr,
+                 const CaptureSettings* captureSettings = nullptr);
```
toloudis, line 102: the renderer can now be initialized with CaptureSettings, so that it can determine on its own whether a render is completed or not. (In particular, CaptureSettings holds on to a RenderDuration object)

agave_app/renderer.h
```diff
@@ -93,30 +145,20 @@ class Renderer
...
-    }
+  int32_t m_width, m_height;
 
-    QString m_name;
-    int m_start;
-    int m_end;
-  };
+  QElapsedTimer m_time;
 
-  QList<SceneDescription> m_scenes;
```
toloudis, line 115: this SceneDescription stuff was just dead unused code, so removing it.

agave_app/renderer.cpp
```diff
@@ -242,6 +273,12 @@ Renderer::processRequest()
     // QMetaObject::invokeMethod(
     //  renderDialog, [=]() { /* ... onRenderRequestProcessed(lastReq, img); ... */ }, Qt::QueuedConnection);
     emit requestProcessed(lastReq, img);
+
+    if (m_streamMode && !shouldContinue()) {
+      m_frameIterations = 0;
+      m_frameTimeSeconds = 0.0f;
+      emit frameDone(img);
+    }
```
toloudis, line 281: this is the check that sends a new message back to the main thread that a rendered frame was completed, so the main thread can update UI progress bars and save the image to disk.

agave_app/renderDialog.cpp
```diff
@@ -791,6 +798,92 @@ RenderDialog::getFullSavePath()
   return saveFilePath;
 }
 
+void
+RenderDialog::onFrameDone(QImage image)
```
toloudis, line 802: most of the body of this function used to be in onRenderRequestProcessed, where we would try to figure out if the frame was completed or not.  With this message handler, it is more definitive.

interim17 (approved):
Appreciate you walking me through this yesterday!

LGTM with the caveat that I haven't built or installed this branch, I've only read through the code.

toloudis:
Big apology, I merged a pull request into this branch which created extra changes in the diff.  Let me know if it's bothersome and I'll fix it. But since we already discussed the PR in meeting, maybe it will be ok.

frasercl (approved):

agave_app/renderer.cpp
```diff
@@ -143,9 +147,33 @@ Renderer::addRequest(RenderRequest* request)
...
+    if (m_myVolumeData.m_captureSettings->renderDuration.durationType == SAMPLES) {
+      if (m_frameIterations < m_myVolumeData.m_captureSettings->renderDuration.samples) {
+        return true;
+      }
+      return false;
+    } else if (m_myVolumeData.m_captureSettings->renderDuration.durationType == TIME) {
+      if (m_frameTimeSeconds < m_myVolumeData.m_captureSettings->renderDuration.duration) {
+        return true;
+      }
+      return false;
```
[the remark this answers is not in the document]
toloudis: hm, yeah why not!

Issue #208, opened 2024-08-01
load tiff time sequences


toloudis:
## Use Case

add a loading option of some sort to figure out that a tiff is part of a time sequence.
discover where the varying "t" value is in the filenames.

## Solution

idk, figure it out somehow with string processing and directory walk the list of filenames.

## Alternatives

maybe let user give hint about the filename structure?

toloudis:
closed by #211

PR #211, opened 2024-08-13
Feature/tiff image sequence

toloudis:
Address #208 

Add a new loader that can handle image sequences of tiffs.
Add a checkbox to the loading dialog to indicate user wishes to load a time sequence. This requires that we use Qt's file dialog rather than native file dialog, in order to be able to add new widgets.

How it works:
User may check the new Image Sequence checkbox in the file loading dialog.

If checked, and the image is a tiff file, then the whole contents of the directory (all *.tif files) will be assumed to be a time sequence of volumes when sorted alphanumerically by name.  No other checks will be done to validate files until they are loaded by moving the time slider.

If checked, and the image is not a tiff, then single-file volume loading will be used.
If unchecked, then single-file volume loading will be used.


renderlib/io/FileReaderImageSequence.cpp
```diff
@@ -0,0 +1,65 @@
...
+  std::vector<std::string> files;
+  std::filesystem::path directory = fpath.parent_path();
+  std::filesystem::path extension = fpath.extension();
+
+  for (const auto& entry : std::filesystem::directory_iterator(directory)) {
+    if (entry.is_regular_file() && entry.path().extension() == extension) {
+      files.push_back(entry.path().string());
+    }
+  }
+  std::sort(files.begin(), files.end());
```
rugeli, line 22: Sorting the tiff files before returning is a smart approach!

rugeli (approved):
The `isImageSequence` check looks solid and the logic in the changes is clear, too. LGTM!

ascibisz (approved):

PR #212, opened 2024-08-15
add support for S3 and gc urls to load zarr

toloudis:
Review size: small/med

Allow "Open URL" to load urls to Zarr data that start with "s3:" or "gc:".

A big piece of this code change is just moving some string utility functions into their own shared module (without any modifications).

Otherwise the key implementation is in FileReaderZarr.cpp.  It splits up the url and tells `tensorstore` what kind of underlying "kvstore driver" to use.

### DIFF
--- a/agave_app/loadDialog.cpp
+++ b/agave_app/loadDialog.cpp
@@ -32,7 +32,7 @@ LoadDialog::LoadDialog(std::string path, const std::vector<MultiscaleDims>& dims
   std::string title;
   QString pathString = QString::fromStdString(path);
   QString extension = pathString.sliced(pathString.lastIndexOf("."));
-  if (pathString.indexOf("http") == 0) {
+  if ((pathString.indexOf("http") == 0) || (pathString.indexOf("s3:") == 0) || (pathString.indexOf("gs:") == 0)) {
     title = "URL Load Settings";
   } else if (extension.indexOf(".zarr") == 0) {
     title = "Directory Load Settings";
--- a/renderlib/StringUtil.cpp
+++ b/renderlib/StringUtil.cpp
@@ -1,0 +1,76 @@
+#include "StringUtil.h"
+
+#include <stringstream>
+
+// TODO: move into a String Utils
+static std::string
+trim(const std::string& str, const std::string& whitespace)
+{
+  const auto strBegin = str.find_first_not_of(whitespace);
+  if (strBegin == std::string::npos)
+    return ""; // no content
+
+  const auto strEnd = str.find_last_not_of(whitespace);
+  const auto strRange = strEnd - strBegin + 1;
+
+  return str.substr(strBegin, strRange);
+}
+
+// TODO: move into a String Utils
+static bool
+startsWith(const std::string mainStr, const std::string toMatch)
+{
+  // std::string::find returns 0 if toMatch is found at starting
+  if (mainStr.find(toMatch) == 0)
+    return true;
+  else
+    return false;
+}
+
+// TODO: move into a String Utils
+static bool
+endsWith(std::string const& value, std::string const& ending)
+{
+  if (ending.size() > value.size())
+    return false;
+  return std::equal(ending.rbegin(), ending.rend(), value.rbegin());
+}
+
+// TODO: move into a String Utils
+static void
+split(const std::string& s, char delim, std::vector<std::string>& elems)
+{
+  std::stringstream ss;
+  ss.str(s);
+  std::string item;
+  while (std::getline(ss, item, delim)) {
+    elems.push_back(item);
+  }
+}
+
+// multi lines split by newline
+// each line split by =
+std::map<std::string, std::string>
+splitToNameValuePairs(const std::string& s)
+{
+  std::vector<std::string> sl;
+  split(s, '\n', sl);
+
+  // split each string into name/value pairs,
+  // then look up as a map.
+  std::map<std::string, std::string> pairs;
+  for (int i = 0; i < sl.size(); ++i) {
+    std::vector<std::string> namevalue;
+    split(sl[i], '=', namevalue);
+    if (namevalue.size() == 2) {
+      pairs[namevalue[0]] = namevalue[1];
+    } else if (namevalue.size() == 1) {
+      pairs[namevalue[0]] = "";
+    } else {
+      // on error return empty map.
+      LOG_ERROR << "Unexpected name/value pair: " << sl[i];
+      return std::map<std::string, std::string>();
+    }
+  }
+  return pairs;
+}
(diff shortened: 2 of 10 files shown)


agave_app/loadDialog.cpp
```diff
@@ -32,7 +32,7 @@ LoadDialog::LoadDialog(std::string path, const std::vector<MultiscaleDims>& dims
   std::string title;
   QString pathString = QString::fromStdString(path);
   QString extension = pathString.sliced(pathString.lastIndexOf("."));
-  if (pathString.indexOf("http") == 0) {
+  if ((pathString.indexOf("http") == 0) || (pathString.indexOf("s3:") == 0) || (pathString.indexOf("gs:") == 0)) {
```
ShrimpCryptid, line 35: It says `gc:` in the description for the PR, but `gs:` here? Which one is right?

renderlib/StringUtil.cpp
```diff
@@ -0,0 +1,74 @@
...
+  else
+    return false;
+}
+
+bool
+endsWith(std::string const& value, std::string const& ending)
+{
+  if (ending.size() > value.size())
+    return false;
+  return std::equal(ending.rbegin(), ending.rend(), value.rbegin());
```
ShrimpCryptid, line 35: Instead of searching at `value.rbegin()`, should this start at an offset `ending.size` characters from the end of `value`?

Something like
`return std::equal(ending.rbegin(), ending.rend(), value.rend() - (ending.rend() - ending.rbegin()))`

renderlib/io/FileReaderZarr.cpp
```diff
@@ -16,16 +17,62 @@
...
+        path += "/";
+      }
+      path = path + subpath;
+      if (!endsWith(path, "/")) {
+        path += "/";
+      }
+      return { { "driver", "gcs" }, { "bucket", bucket }, { "path", path } };
+    } else {
+      return { { "driver", "http" }, { "base_url", filepath }, { "path", subpath } };
+    }
```
ShrimpCryptid, line 75: Looks good! I'd love to see some unit tests for this though

ShrimpCryptid (commented):
Couple small comments. I think the string methods would really benefit from unit tests!

renderlib/StringUtil.cpp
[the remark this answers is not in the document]
toloudis: rbegin and rend are iterators that start from the END of the string. The `r` means reverse.  So rbegin starts at the end and iterates backward to rend (which represents the beginning!)  And in the flavor of std::equal that takes 3 args, the second range is taken to have the same number of elements as the first range.  In this case that means the length of the `ending` string.

renderlib/io/FileReader.cpp
```diff
@@ -42,6 +42,10 @@ FileReader::getReader(const std::string& filepath, bool isImageSequence)
     return new FileReaderImageSequence(filepath);
   } else if (filepath.find("http") == 0) {
     return new FileReaderZarr(filepath);
+  } else if (filepath.find("s3:") == 0) {
+    return new FileReaderZarr(filepath);
+  } else if (filepath.find("gs:") == 0) {
+    return new FileReaderZarr(filepath);
```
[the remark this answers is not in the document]
toloudis: Might as well..  in this case the expression is so short, it makes little difference but why not!

toloudis:
> Couple small comments. I think the string methods would really benefit from unit tests!

Good idea, I can throw some in!

renderlib/io/FileReader.cpp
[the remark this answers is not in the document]
toloudis: also bear in mind that this StringUtils was already working code in the places where it was used.  I am just moving it into a common file for this pr.

test/test_stringUtil.cpp
```diff
@@ -0,0 +1,108 @@
...
+    split("hello world", 'x', elems);
+    REQUIRE(elems.size() == 1);
+    REQUIRE(elems[0] == "hello world");
+
+    elems.clear();
+    split("hello world", 'o', elems);
+    REQUIRE(elems.size() == 3);
+    REQUIRE(elems[0] == "hell");
+    REQUIRE(elems[1] == " w");
+    REQUIRE(elems[2] == "rld");
```
ShrimpCryptid, line 61: hell world 😱

test/test_stringUtil.cpp
```diff
@@ -0,0 +1,108 @@
...
+    REQUIRE(elems.size() == 2);
+    REQUIRE(elems[0] == "name");
+    REQUIRE(elems[1] == "value");
+
+    elems.clear();
+    split("name=value\nname2=value2\n", '\n', elems);
+    REQUIRE(elems.size() == 2);
+    REQUIRE(elems[0] == "name=value");
+    REQUIRE(elems[1] == "name2=value2");
+  }
```
ShrimpCryptid, line 74: Hmm, this is interesting to me. I think in JS the expected behavior is `["name=value", "name2=value2", ""]` where the ending `\n` would cause an empty string at the end? Any reason to just outright discard it?

ShrimpCryptid: Especially since it looks like you're adding a check for empty strings in `splitToNameValuePairs`...

test/test_stringUtil.cpp
```diff
@@ -0,0 +1,108 @@
...
+    REQUIRE(pairs.size() == 2);
+    REQUIRE(pairs["name"] == "value");
+    REQUIRE(pairs["name2"] == "value2");
+
+    pairs = splitToNameValuePairs("name=value\nname2=value2\nname3=value3\n\n");
+    REQUIRE(pairs.size() == 3);
+    REQUIRE(pairs["name"] == "value");
+    REQUIRE(pairs["name2"] == "value2");
+    REQUIRE(pairs["name3"] == "value3");
+  }
```
ShrimpCryptid, line 107: Maybe add one more test where there's multiple newlines between two elements?

ShrimpCryptid (commented):
Ooh, these unit tests are great, thank you! I added some last nit comments, will approve once you respond!

test/test_stringUtil.cpp
toloudis: I added an extra test that shows empty string.  This is just how my implementation works.  It's using std::getline with the specified delimiter.  When the delimiter is at the end of the string, you don't get any tokens after that.

test/test_stringUtil.cpp
toloudis: ok done!

agave_app/loadDialog.cpp
meganrm, line 35: Is there a possibility there will be more of these in the future? Just thinking this looks like a place for a "isURL" function with approved prefixes

toloudis: I'm not aware of any yet but I agree this is about at the limit where if any new one comes up it will get refactored.

meganrm (approved):

ShrimpCryptid (approved):

PR #220, opened 2024-09-26
Read zarrv3 with OME-NGFF 0.5 json

toloudis:
Add the ability to load from the latest OME-NGFF conversion test datasets. 
Example data is here:  https://samples-viewer--ome2024-ngff-challenge.netlify.app/?csv=https://raw.githubusercontent.com/will-moore/ome2024-ngff-challenge/samples_viewer/samples/ngff_samples.csv

This is a very barebones first try but it seems to load basically all of the files.  It doesn't try to load any extra metadata yet.  Thankfully the `axes` and `datasets` follow basically the same schema.  If that schema diverges then this code has to be refactored again. 

The big assumption I am making is that OME-NGFF 0.4 is always zarrv2 and if we find a zarrv3 zarr.json file, then I am assuming a OME-NGFF 0.5 json file.   A further PR can add some extra logic and validation around this but for now things seem to be working.  Anything outside this assumption will error out in some way.


renderlib/io/FileReader.cpp
```diff
@@ -55,6 +55,13 @@ FileReader::getReader(const std::string& filepath, bool isImageSequence)
   } else if (extstr == ".zarr") {
     return new FileReaderZarr(filepath);
   }
+  // if it's a directory, and contains the string zarr anywhere, we assume it's a zarr
+  else if (std::filesystem::is_directory(filepath)) {
+    if (filepath.find("zarr") != std::string::npos) {
+      return new FileReaderZarr(filepath);
+    }
+  }
+
```
toloudis, line 64: this change is a throw-in to allow loading zarrs from local directories when the directory name doesn't end in .zarr.  (Some conversion tools put the zarr in a subdirectory with a name like `my_file.zarr/0/`.

renderlib/io/FileReaderZarr.cpp
```diff
@@ -123,7 +123,35 @@ FileReaderZarr::jsonRead(const std::string& zarrurl)
...
+    hasV3 = true;
+  } else {
+    attrs = tryReadJson(zarrurl, ".zattrs");
+  }
+  if (attrs.is_object() && !attrs.empty()) {
+    LOG_DEBUG << (hasV3 ? "Zarr v3" : "Zarr v2");
+  } else {
+    LOG_ERROR << "Failed to open either a .zattrs or a zarr.json from the specified location " << zarrurl;
+    return ::nlohmann::json::object_t();
+  }
```
ShrimpCryptid, line 151: Nit: I feel like the control flow on this is reads a little weird? It might be nicer to have the V3 return early instead of rejoining and doing separate logic with a flag.

renderlib/io/FileReaderZarr.cpp
```diff
@@ -143,11 +171,57 @@ FileReaderZarr::getChannelNames(const std::string& filepath)
...
+
+nlohmann::json
+FileReaderZarr::getMultiscales(nlohmann::json attrs)
+{
+  nlohmann::json multiscales;
+  if (m_zarrVersion == 3) {
+    auto attributes = attrs["attributes"];
+    if (attributes.is_null()) {
+      LOG_ERROR << "No attributes found in zarr.json";
+      return multiscales;
```
ShrimpCryptid, line 183: Is `multiscales` a nullptr in this return statement?

renderlib/io/FileReaderZarr.cpp
```diff
@@ -143,11 +171,57 @@ FileReaderZarr::getChannelNames(const std::string& filepath)
...
+      LOG_ERROR << "No attributes.ome found in zarr.json";
+      return omero;
+    }
+    omero = ome["omero"];
+  } else {
+    omero = attrs["omero"];
+  }
+
+  return omero;
+}
```
ShrimpCryptid, line 218: Nit: Move before usage on line 163?

ShrimpCryptid (approved):
Left a few code style comments but nothing blocking!

I haven't dug into the OME ZARR spec so I can't comment much there. Is there a good place in the docs that defines the two different schemas?

renderlib/io/FileReaderZarr.cpp
```diff
@@ -330,7 +411,7 @@ FileReaderZarr::loadMultiscaleDims(const std::string& filepath, uint32_t scene)
         auto path = dataset["path"];
         if (path.is_string()) {
           std::string pathstr = path;
-          auto result = tensorstore::Open({ { "driver", "zarr" },
+          auto result = tensorstore::Open({ { "driver", tensorstoreZarrDriverName() },
```
ShrimpCryptid, line 414: So tenserstore has its own implementation for the drivers for both Zarr v2 and v3?

renderlib/command.cpp
```diff
@@ -601,14 +601,12 @@ SetTimeCommand::execute(ExecutionContext* c)
...
     GradientData& lutInfo = c->m_appScene->m_material.m_gradientData[i];
     lutInfo.convert(c->m_appScene->m_volume->channel(i)->m_histogram, image->channel(i)->m_histogram);
-
     image->channel(i)->generateFromGradientData(lutInfo);
   }
 
   // now we're ready to lose the old channel histograms
   c->m_appScene->m_volume = image;
 
-  c->m_renderSettings->m_DirtyFlags.SetFlag(VolumeDirty);
```
toloudis, line 611: this line removal is a bonus bug fix for rendering time series. Setting VolumeDirty is completely reallocating texture and resetting too much stuff, when VolumeDataDirty is all that is needed.

renderlib/io/FileReaderZarr.cpp
```diff
@@ -123,7 +123,35 @@ FileReaderZarr::jsonRead(const std::string& zarrurl)
...
+  // try zarr.json first (indicating Zarr v3)
+  // and then try .zattrs (indicating Zarr v2)
+  nlohmann::json attrs = tryReadJson(zarrurl, "zarr.json");
+  if (attrs.is_object() && !attrs.empty()) {
+    hasV3 = true;
+  } else {
+    attrs = tryReadJson(zarrurl, ".zattrs");
+  }
+  if (attrs.is_object() && !attrs.empty()) {
+    LOG_DEBUG << (hasV3 ? "Zarr v3" : "Zarr v2");
```
rugeli, line 147: This log seem a bit unintuitive to me as a non-frequent user, especially this check is identical to the one on L141. Does it indicate that it's loading V3/V2 Zarr metadata from the given `zarrurl`?

rugeli (approved):
Overall lgtm!

renderlib/io/FileReaderZarr.cpp
toloudis: I see your point. that might be a bit more readable, I'll type it in and see how it looks

renderlib/io/FileReaderZarr.cpp
toloudis: basically, yeah, it's a default-initialized empty nlohmann::json  object. (this is an error)

renderlib/io/FileReaderZarr.cpp
toloudis: ok

renderlib/io/FileReaderZarr.cpp
toloudis: the value of attrs has changed here because we tried to load json from a different filename (zarr.json vs .zattrs). that's why the check happens a second time whether or not the first time failed.
After we have tried twice, we know all we need to know to continue.
Based on Peyton's comment, I think I will clean this up a little bit.

PR #225, opened 2024-11-10
Feature/refactor selection manipulator

toloudis:
Review time: medium.

This is preliminary work before adding a user clip plane that we want to let the user interactively position and rotate.  This PR should cause no behavior changes to the app.

Previously the interactive rotation gizmo (aka RotateTool) was only used for the scene area light source.  Several things about this were hardcoded.

The key change here is to add the notion of a selection object to the scene.  The selection object is a generic SceneObject which has a transform that can be manipulated using the RotateTool.

As a side effect of this change, the way light sources are initialized had to be changed.  In addition I added some convenience methods for getting the AreaLight or the SphereLight without having to remember what integer index it is.


renderlib/AppScene.h
```diff
@@ -45,41 +46,45 @@ class Lighting
...
+
+    m_Lights[i] = new Light(light);
+    m_sceneLights[i] = new SceneLight(m_Lights[i]);
   }
 
-  Light m_Lights[MAX_NO_LIGHTS];
+  Light* m_Lights[MAX_NO_LIGHTS];
   int m_NoLights;
-  SceneLight m_sceneLights[MAX_NO_LIGHTS];
+  SceneLight* m_sceneLights[MAX_NO_LIGHTS];
```
toloudis, line 87: a key change here (and in the lines above) is that the light sources are now stored and handled as pointers instead of raw objects.   For convenience, in some places we access them as pointers and in other places as references.  This can be neatened up in another PR (maybe using smart pointers) but doesn't clutter this change set too much.

renderlib/SceneView.h
```diff
@@ -58,5 +59,8 @@ struct SceneView
   Scene* scene = nullptr;
   RenderSettings* renderSettings = nullptr;
 
-  bool anythingActive() const { return scene != nullptr; }
+  // weak ptr! must not outlive the objects it points to.
+  SceneObject* selection = nullptr;
```
toloudis, line 63: key change: this selection object will be able to be affected by transform manipulators

agave_app/GLView3D.cpp
```diff
@@ -309,14 +309,14 @@ GLView3D::toggleAreaLightRotateControls()
 {
   // toggle rotate tool
   if (m_areaLightMode == AREALIGHT_MODE::NONE || m_areaLightMode == AREALIGHT_MODE::TRANS) {
-    m_viewerWindow->showAreaLightGizmo(true);
+    m_viewerWindow->select(m_viewerWindow->sceneView.scene->SceneAreaLight());
```
toloudis, line 312: key change: now we just tell the viewerWindow what the selected SceneObject is.  If the object provides a Tool for drawing itself when selected (e.g. a wireframe box), then that will automatically be added.

renderlib/CCamera.h
toloudis, line 1: the changes in this file are just minor cleanup and resolving one uninitialized data member.

renderlib/CCamera.h
```diff
@@ -352,6 +352,7 @@ class CCamera
     m_Dirty = true;
     m_Projection = PERSPECTIVE;
     m_OrthoScale = DEF_ORTHO_SCALE;
+    m_AreaPixel = m_Film.m_Resolution.GetAspectRatio() / (m_Focus.m_FocalDistance * m_Focus.m_FocalDistance);
```
ShrimpCryptid, line 355: What does `m_AreaPixel` represent? The area of a pixel at the focal distance?

renderlib/Origins.cpp
```diff
@@ -10,14 +10,18 @@ Origins::update(SceneView& scene)
...
-    m_origins = { lt.m_transform };
-    // we want the rotate manipulator to be centered at the target of the light
-    m_origins[0].m_center = lt.m_light->m_Target;
+    m_origins = { obj->m_transform };
+
+    // arealight special case:
+    // we want the rotate manipulator to be centered at the target of the light.
+    // This could be a virtual SceneObject method like getManipulatorCenter()
+    Light& lt = scene.scene->AreaLight();
+    m_origins[0].m_center = lt.m_Target;
```
ShrimpCryptid, line 24: Should `0` be a hardcoded value here?

renderlib/ViewerWindow.cpp
```diff
@@ -56,17 +50,27 @@ ViewerWindow::setSize(int width, int height)
...
-  if (show) {
-    m_tools.push_back(m_areaLightTool);
-  } else {
-    m_tools.erase(std::remove(m_tools.begin(), m_tools.end(), m_areaLightTool), m_tools.end());
+
+  // deselect the current selection
+  if (sceneView.selection) {
+    ManipulationTool* tool = sceneView.selection->getSelectedTool();
+    if (tool) {
+      m_tools.erase(std::remove(m_tools.begin(), m_tools.end(), tool), m_tools.end());
```
ShrimpCryptid, line 64: Can you explain what this line is doing? I see that it's basically unchanged from the diff but I wasn't sure how the `remove` and `erase` are working.

ShrimpCryptid (commented):
Leaving some questions!

renderlib/AppScene.h
```diff
@@ -95,6 +100,15 @@ class Scene
...
 
+  // convenience functions
+  // For now, this must match the order in which the lights were added, in initLights
+  static constexpr int SphereLightIndex = 0;
+  Light& SphereLight() const { return *m_lighting.m_Lights[SphereLightIndex]; }
+  SceneLight* SceneSphereLight() const { return m_lighting.m_sceneLights[SphereLightIndex]; }
+  static constexpr int AreaLightIndex = 1;
+  Light& AreaLight() const { return *m_lighting.m_Lights[AreaLightIndex]; }
+  SceneLight* SceneAreaLight() const { return m_lighting.m_sceneLights[AreaLightIndex]; }
+
```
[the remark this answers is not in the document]
toloudis: yeah currently one could do that.  It could happen if data were saved out badly, or corrupted after being saved, but I think that's the only way.   I could nail things down even more, of course.

renderlib/CCamera.h
toloudis: yeah this is just a throw-in for a lint warning I was seeing about uninitialized data.  I don't think the value is even used anywhere currently.  could be used for depth-of-field blurring or adaptive sampling bla bla bla.

renderlib/Origins.cpp
toloudis: The origins class is "prepared" to hold an array of objects but currently only the first object is ever active.  The idea being that you could have multiple objects selected and then rotate or translate all of them via the Origins class.

renderlib/ViewerWindow.cpp
toloudis: Yeah this is a bit funky but a well established pattern.  The std::remove actually shifts the queried object (`tool`) to the end of the vector `m_tools` and returns an iterator to the start of the removed items.  Then erase actually drops everything from where that iterator starts.  
see here:
https://stackoverflow.com/questions/347441/how-can-you-erase-elements-from-a-vector-while-iterating

renderlib/AppScene.h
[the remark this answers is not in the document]
toloudis: Honestly the way the light sources are handled right now is still sloppy in part because SceneObject (and SceneLight) is a new thing to this codebase.  A future PR will need to clean up the creation, deletion and handling of SceneObjects overall.

renderlib/ViewerWindow.cpp
ShrimpCryptid: Gotcha, thank you!

ShrimpCryptid (approved):
LGTM! I think I had a hard time fully understanding all the changes just because they were so spread out across files + unfamiliar codebase, but thanks for answering the questions I had.

renderlib/AxisHelperTool.cpp
```diff
@@ -1,5 +1,7 @@
 #include "AxisHelperTool.h"
 
+#include "AppScene.h"
+
```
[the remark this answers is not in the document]
toloudis: I honestly can't remember right now but it could be because of changes in AppScene or things that AppScene includes, too.
Sometimes it's easy to just include extra header files that never get used.  The compiler doesn't really tell you about that - you have to use static analysis tools to clean up extra unnecessary includes.

PR #229, opened 2024-11-19
Feature/refactor for clip plane

toloudis:
This is a followup refactor on top of #225 

The goal of this change set is to finish decoupling selection from manipulation, and prepare the code to have the clip plane added.

Key changes:
* remove some special-case code for light source manipulation. For example, renaming AREA_LIGHT_MODE to MANIPULATOR_MODE.
* light source (or any object) must now be explicitly selected (in code) before it can be rotated
* move Transform3d out of Object3d and into MathUtils. Improve Transform3d::getMatrix
* get ready to allow both rotation and translation modes 

The next PR will actually add user clipping plane

### DIFF
--- a/agave_app/GLView3D.cpp
+++ b/agave_app/GLView3D.cpp
@@ -362,7 +362,7 @@ GLView3D::toggleTranslateControls()
   if (!m_viewerWindow->sceneView.getSelectedObject()) {
     setManipulatorMode(MANIPULATOR_MODE::NONE);
   }
-  // toggle rotate tool
+  // toggle translate tool
   else if (m_manipulatorMode == MANIPULATOR_MODE::TRANS) {
     setManipulatorMode(MANIPULATOR_MODE::NONE);
   } else {


CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ project(
 )
 
 set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
toloudis, line 23: this is purely an improvement for vscode performance issues. Has no effect on builds.

pyrightconfig.json
toloudis, line 1: This file is added purely for vscode performance reasons and has no effect on any builds or behavior of agave.

renderlib/renderlib.h
toloudis, line 1: The includes for some of the Qt classes in renderlib were moved around to make the builds a tiny bit faster. Qt includes can be a bit more heavyweight.

ShrimpCryptid (approved):
Had just the one question about how the emit/selection mechanism works, thank you for answering that! LGTM

interim17 (approved):
LGTM

PR #236, opened 2025-01-18
Update Qt to 6.8

toloudis:
keeping up with current Qt.
This revealed bug with how we use the QOpenGLWidget which is fixed here.


rugeli (approved):
It seems there's a failing workflow job that we need to address, otherwise the version updates look good to me!

agave_app/GLView3D.cpp
```diff
@@ -164,9 +164,9 @@ GLView3D::paintGL()
     return;
   }
 
-  makeCurrent();
+  // makeCurrent();
   m_viewerWindow->redraw();
-  doneCurrent();
+  // doneCurrent();
```
rugeli, line 169: Are these commented out for testing or is this part of the code change?

toloudis:
> It seems there's a failing workflow job that we need to address, otherwise the version updates look good to me!

Sorry, the failing workflow job is consistent across several PRs now and is due to a broken dependency build on github-actions.  I'm working on that but ask that you forgive it for these PRs.

agave_app/GLView3D.cpp
toloudis: It's part of the code change. I'll delete them.

ascibisz (approved):

PR #237, opened 2025-01-20
Feature/refactor shaders

toloudis:
Time to review: pretty quick

This code change appears to add lots of lines but it's because the shader sources are doubly added to the repo.
There should be no resulting changes in functionality in AGAVE from this PR.

Shaders were previously stored inline in code as string literals.
This code change is to allow improved shader editing experience for devs, and also to prepare for porting to other apis like vulkan or wgpu.

* Move all shader source code into .vert and .frag files.  These files are now just glsl code and are the source of truth.  Because they are standalone they can be properly syntax-highlighted etc.
* Generate c++ source files from the vert and frag files using make_shaders.py.  Currently this has to be done manually.  The generated source files are to be checked in as part of the codebase. (TODO: how to guarantee that they are up-to-date? have cmake generate them all the time?)
* The cmake build then compiles the c++ source files.
* All shader sources are stored in a keyed map object, and then they are retrieved by keyed lookups.  GPU Shaders are still compiled lazily on-demand in the renderers.  This new shader table data structure could evolve into pre-compiling all the shaders at app startup.


rugeli (approved):
Not so familiar with cpp but I have gone over the shader conversion python script and the generated source files look good!

meganrm (approved):

PR #238, opened 2025-02-05
Feature/serialize clip plane

toloudis:
Clip plane gets written to json as : default plane {(0,0,1),0} with a transform (a translation, rotation and scale).  This is how the plane is stored internally, as well.

Clip plane goes through Python api as: plane equation {(nx,ny,nz),d} expressed as normal vector and distance.  This was a simplifying choice to get the Python api to work. 

* FUTURE TODO: A big overhaul of the Python api is in order, to handle transforms more generally since there are now several object types that can be transformed. 
* INTERIM: a smaller PR may follow to allow more intuitive ways to transform the plane via Python.

### DIFF
--- a/renderlib/StringUtil.cpp
+++ b/renderlib/StringUtil.cpp
@@ -2,6 +2,7 @@
 
 #include "Logging.h"
 
+#include <regex>
 #include <sstream>
 
 std::string
@@ -73,4 +74,10 @@ splitToNameValuePairs(const std::string& s)
     }
   }
   return pairs;
+}
+
+std::string
+escapePath(const std::string& path)
+{
+  return std::regex_replace(path, std::regex("\\\\"), "\\\\");
 }
--- a/renderlib/io/FileReaderZarr.cpp
+++ b/renderlib/io/FileReaderZarr.cpp
@@ -533,19 +533,18 @@ FileReaderZarr::loadFromFile(const LoadSpec& loadSpec)
     tsdim++;
 
     static constexpr tensorstore::DimensionIndex kNumDims = 5;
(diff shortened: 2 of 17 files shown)


renderlib/io/FileReaderZarr.cpp
```diff
@@ -533,18 +533,19 @@ FileReaderZarr::loadFromFile(const LoadSpec& loadSpec)
     tsdim++;
 
     static constexpr tensorstore::DimensionIndex kNumDims = 5;
-    const tensorstore::Index shapeToLoad[kNumDims] = { 1, 1, dims.sizeZ, dims.sizeY, dims.sizeX };
+    // const tensorstore::Index shapeToLoad[kNumDims] = { 1, 1, dims.sizeZ, dims.sizeY, dims.sizeX };
```
ShrimpCryptid, line 536: Remove?

renderlib/StringUtil.cpp
```diff
@@ -74,3 +75,9 @@ splitToNameValuePairs(const std::string& s)
   }
   return pairs;
 }
+
+std::string
+escapePath(const std::string& path)
+{
+  return std::regex_replace(path, std::regex("\\\\"), "\\\\");
```
ShrimpCryptid, line 82: Is this replacing `\\` with `\\\\`? Do you also need to worry at all about `\` characters?

ShrimpCryptid (approved):

renderlib/StringUtil.cpp
toloudis: I realize it looks odd but the regex string itself has to be escaped, because of what backslash means in a regex.  so the regex is actually `\\` which means to find single backslash.  And I'm replacing it with `\\` a double backslash.  This function is used when I need to output a filepath as a python string to construct a python script..

interim17 (approved):
LGTM

PR #239, opened 2025-02-09
Feature/serialize color map

toloudis:
Color Maps need to be written out for json save/load, and also Python script export. 

Currently, there are only static choices of color maps. Therefore the NAME of the color map determines all of it.  However, I also allow all the color ramp gradient stops to be written out as well.  This is a bit of future-proofing for allowing users to completely customize the color ramp.

The biggest change is a refactor which encapsulates the color maps into a class called ColorRamp, and also moved its storage out of the Channelu16 class and into the VolumeDisplay.


agave_app/SerializeV1.h
```diff
@@ -99,6 +99,15 @@ struct LutParams_V1
   }
   NLOHMANN_DEFINE_TYPE_INTRUSIVE(LutParams_V1, window, level, isovalue, isorange, pctLow, pctHigh, controlPoints, mode)
 };
+
+struct ColorMap
+{
+  std::string name = "none";
+  std::vector<ControlPointSettings_V1> stops;
+  bool operator==(const ColorMap& other) const { return name == other.name && stops == other.stops; }
+  NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(ColorMap, name, stops)
```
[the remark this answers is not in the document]
toloudis: this macro is from the third-party json library.  It lets you decorate all the member variables of a struct so that it can be read from and written to json.  "WITH_DEFAULT" allows you to add new fields later down the road, and old json versions of the struct will still load.

renderlib/Colormap.cpp
```diff
@@ -553,11 +554,52 @@ static std::vector<std::pair<std::string, const std::vector<ColorControlPoint>>>
...
+    // the name is actually in our list:
+    if (gspec.m_name == name) {
+      return gspec;
+    }
+  }
+
+  LOG_ERROR << "Unknown colormap name: " << name << ". Falling back to no colormap.";
+  // use "none" as a special case when the map is not found.
+  return colormapFromName(NO_COLORMAP_NAME);
+}
```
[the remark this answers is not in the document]
toloudis: 2.  It is called elsewhere by the upper levels of the app to build the UI dropdown box for selecting the colormap.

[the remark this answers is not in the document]
toloudis: 1. I considered indexing directly by just knowing that NO_COLORMAP_NAME is always index 0 anyway, but I sort of wanted it to be independent of that, so as to be more immune to future changes. 
The recursion will only happen once, because I am darn sure that the colormap exists.    I suppose for robustness I could assert somehow that NO_COLORMAP_NAME exists before making the recursive call. 
I guess you're right in that I could check for NO_COLORMAP_NAME while looping and remember its index.

renderlib/Colormap.h
```diff
@@ -60,17 +60,76 @@ struct ColorControlPoint
...
+  // Move assignment operator
+  ColorRamp& operator=(ColorRamp&& other) noexcept
+  {
+    if (this != &other) {
+      m_name = std::move(other.m_name);
+      m_stops = std::move(other.m_stops);
+      m_colormap = std::move(other.m_colormap);
+    }
+    return *this;
+  }
```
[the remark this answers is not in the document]
toloudis: there is a suggested change but I don't see the diff.

Yes, std::move invalidates the object being moved.  The move assignment operator and move copy constructor are there to optimize situations where you have a temporary variable that only exists to get passed into a new var, and would ordinarily be fully copied an extra time.  The temp variable is going away anyway and is not going to be used again.

See "rule of five" on this page: https://en.cppreference.com/w/cpp/language/rule_of_three
Admittedly this rule has not been followed in many classes in this app - I am doing this in part as an exercise to start getting used to it for data types that might get passed around a lot, that hold blocks of more than a few bytes.

agave_app/AppearanceSettingsWidget.cpp
```diff
@@ -57,22 +57,26 @@ makeGradientCombo()
...
+    if (gspec.m_name == "Labels") {
+      // special case for Labels
+      brush.setColor(QColor(255, 255, 255));
+      brush.setStyle(Qt::SolidPattern);
+      itemText = gspec.m_name.c_str();
+    } else {
+      brush.setStyle(Qt::LinearGradientPattern);
+    }
+    cb->addItem(itemText, QVariant(gspec.m_name.c_str()));
+    cb->setItemData(index, QVariant(gspec.m_name.c_str()), Qt::ToolTipRole);
```
toloudis, line 76: the above code is just here to set up the dropdown box for the UI, given a list of color gradients.

agave_app/AppearanceSettingsWidget.cpp
```diff
@@ -1078,7 +1071,12 @@ QAppearanceSettingsWidget::onNewImage(Scene* scene)
     // init
     this->OnOpacityChanged(i, scene->m_material.m_opacity[i]);
 
+    // get color ramp from scene
+    const ColorRamp& cr = scene->m_material.m_colormap[i];
     QComboBox* gradients = makeGradientCombo();
+    int idx = gradients->findData(QVariant(cr.m_name.c_str()), Qt::UserRole);
+    LOG_DEBUG << "Found gradient " << idx << " (" << cr.m_name << ") for channel " << i;
+    gradients->setCurrentIndex(idx);
```
toloudis, line 1079: the above code makes sure that on loading from a saved json session, that we initialize the incoming colormaps in the dropdown for each channel

agave_app/SerializeV1.h
toloudis, line 108: this struct is here in SerializeV1.h because it is used exclusively for saving/loading from file as json.  The data structures for file i/o are separated from the internal structures of the app.  Even if they tend to be the same, sometimes there are reasons why they can't be.

agave_app/ViewerState.cpp
```diff
@@ -236,6 +236,15 @@ stateToPythonScript(const Serialize::ViewerState& s)
...
     }
+    std::vector<float> v;
+    for (auto p : ch.colorMap.stops) {
+      v.push_back(p.x);
+      v.push_back(p.value[0]);
+      v.push_back(p.value[1]);
+      v.push_back(p.value[2]);
+      v.push_back(p.value[3]);
+    }
+    ss << obj << SetColorRampCommand({ i, ch.colorMap.name, v }).toPythonString() << std::endl;
```
toloudis, line 247: writing the color map as a python command for python script export

agave_app/ViewerState.cpp
```diff
@@ -299,6 +308,21 @@ stateToLoadSpec(const Serialize::ViewerState& state)
   return spec;
 }
 
+ColorRamp
+stateToColorRamp(const Serialize::ViewerState& state, int channelIndex)
```
toloudis, line 312: this is the "input" from file to app.  taking the Serialize::ViewerState structure and translating it to the app internals.

agave_pyclient/agave_pyclient/commandbuffer.py
```diff
@@ -78,6 +78,8 @@
     "SET_FLIP_AXIS": [46, "I32", "I32", "I32"],
     "SET_INTERPOLATION": [47, "I32"],
     "SET_CLIP_PLANE": [48, "F32", "F32", "F32", "F32"],
+    # channel index, colormap name, array of [stop, r, g, b, a]
+    "SET_COLOR_RAMP": [49, "I32", "S", "F32A"],
```
toloudis, line 82: defines the binary footprint of the arguments for this command, details in command.h and command.cpp

renderlib/AppScene.h
```diff
@@ -35,6 +36,7 @@ struct VolumeDisplay
   float m_opacity[MAX_CPU_CHANNELS];
   bool m_enabled[MAX_CPU_CHANNELS];
   float m_labels[MAX_CPU_CHANNELS];
+  ColorRamp m_colormap[MAX_CPU_CHANNELS];
```
toloudis, line 39: one color ramp for each channel, now stored in Scene but formerly were stored in ImageXYZC::Channelu16

renderlib/Colormap.h
```diff
@@ -60,17 +60,76 @@ struct ColorControlPoint
...
 
-uint8_t*
-colormapFromControlPoints(std::vector<ColorControlPoint> pts, size_t length = 256);
-uint8_t*
-modifiedGlasbeyColormap(size_t length = 256);
-uint8_t*
-colormapRandomized(size_t length = 256);
-uint8_t*
-colormapFromColormap(uint8_t* colormap, size_t length = 256);
+class ColorRamp
```
toloudis, line 63: the changes in this file introduce the ColorRamp class which encapsulates a name, the color gradient stops, and a "baked" or cached version of the colormap as an array / lookup table.

renderlib/Fuse.cpp
```diff
@@ -11,8 +12,7 @@
 void
 Fuse::fuse(const ImageXYZC* img,
            const std::vector<glm::vec3>& colorsPerChannel,
-           const GradientData* channelGradientData,
-           const float* channelIsLabels,
+           const VolumeDisplay& volumeDisplay,
```
toloudis, line 15: changes in this file are just because the colormaps, label info, and gradient data are all stored in VolumeDisplay

renderlib/ImageXYZC.h
```diff
@@ -24,11 +24,7 @@ struct Channelu16
 
   Histogram m_histogram;
   float* m_lut;
-  uint8_t* m_colormap;
```
toloudis, line 27: this is where the colormap used to be stored, per-channel

renderlib/command.cpp
```diff
@@ -769,6 +768,39 @@ SetClipPlaneCommand::execute(ExecutionContext* c)
   c->m_renderSettings->m_DirtyFlags.SetFlag(RenderParamsDirty);
 }
 
+void
+SetColorRampCommand::execute(ExecutionContext* c)
```
toloudis, line 772: any new command has to provide a `execute`, `parse` , `write`, and `toPythonString` implementation

renderlib/command.h
```diff
@@ -497,3 +497,14 @@ CMDDECL(SetClipPlaneCommand,
...
+struct SetColorRampCommandD
+{
+  int32_t m_channel;
+  std::string m_name;
+  std::vector<float> m_data;
+};
+CMDDECL(SetColorRampCommand,
+        49,
+        "set_color_ramp",
+        CMD_ARGS({ CommandArgType::I32, CommandArgType::STR, CommandArgType::F32A }));
```
toloudis, line 510: this is how any new command is declared.  The CommandArgTypes have to correspond to the data members in the CommandD struct above.

renderlib/io/FileReaderZarr.cpp
```diff
@@ -341,10 +341,10 @@ FileReaderZarr::loadMultiscaleDims(const std::string& filepath, uint32_t scene)
             LOG_ERROR << "Error: " << result.status();
             LOG_ERROR << "Failed to open store for " << filepath << " :: " << pathstr;
           } else {
-            tensorstore::TensorStore<> store = result.value();
+            auto store = result.value();
+
             tensorstore::DataType dtype = store.dtype();
-            tensorstore::IndexDomainView<> domain = store.domain();
-            auto shape_span = domain.shape();
+            auto shape_span = store.domain().shape();
```
toloudis, line 347: out of scope of this PR, remnant of fiddling with tensorstore build breakage

test/test_commands.cpp
```diff
@@ -461,4 +461,13 @@ TEST_CASE("Commands can write and read from binary", "[command]")
...
   }
+  SECTION("SetColorRampCommand")
+  {
+    SetColorRampCommandD data = { 3, "HELLO_WORLD", { 0.25, 0.5, 0.333, 0.666, 0.999 } };
+    auto cmd = testcodec<SetColorRampCommand, SetColorRampCommandD>(data);
+    REQUIRE(cmd->toPythonString() == "set_color_ramp(3, \"HELLO_WORLD\", [0.25, 0.5, 0.333, 0.666, 0.999])");
+    REQUIRE(cmd->m_data.m_channel == data.m_channel);
+    REQUIRE(cmd->m_data.m_name == data.m_name);
+    REQUIRE(cmd->m_data.m_data == data.m_data);
+  }
```
toloudis, line 472: new commands should also get test cases just to prove they can transmit their data correctly

agave_app/AppearanceSettingsWidget.cpp
toloudis: remove debug log

meganrm (approved):

PR #248, opened 2025-03-06
transfer function flag should be set when channels enabled/disabled

toloudis:
time to review: tiny!

When enabling channels, the transfer function was not being properly updated.  This is a regression from how colormaps were added.

The app has "dirty flags" for both updating the volume data and the transfer function data separately.  When channels are enabled/disabled, both flags need to be set.

### DIFF
--- a/renderlib/graphics/RenderGLPT.cpp
+++ b/renderlib/graphics/RenderGLPT.cpp
@@ -206,7 +206,6 @@ RenderGLPT::doRender(const CCamera& camera)
     uint32_t c0, c1, c2, c3;
     m_scene->getFirst4EnabledChannels(c0, c1, c2, c3);
     m_imgGpu.updateVolumeData4x16(m_scene->m_volume.get(), c0, c1, c2, c3);
-    m_imgGpu.updateLutGPU(m_scene->m_volume.get(), c0, c1, c2, c3, m_scene->m_material);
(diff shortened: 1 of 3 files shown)


renderlib/graphics/RenderGLPT.cpp
```diff
@@ -206,7 +206,6 @@ RenderGLPT::doRender(const CCamera& camera)
     uint32_t c0, c1, c2, c3;
     m_scene->getFirst4EnabledChannels(c0, c1, c2, c3);
     m_imgGpu.updateVolumeData4x16(m_scene->m_volume.get(), c0, c1, c2, c3);
-    m_imgGpu.updateLutGPU(m_scene->m_volume.get(), c0, c1, c2, c3, m_scene->m_material);
```
ShrimpCryptid, line 209: Why was this line removed?

ShrimpCryptid (approved):

frasercl (approved):

renderlib/graphics/RenderGLPT.cpp
toloudis: Because it's in the part of the code that is only supposed to update the volume data.  It was the wrong place to update the LUT on GPU and because of the dirty flags, it is redundant to have it happen here.

PR #253, opened 2025-03-24
add reset button for clip plane

toloudis:
This code change adds a button which resets the clip plane to its initial position, centered at the center of the volume and oriented along +Z.

This is added after testing with Thao who wanted a quick way to get the plane back to its starting point after doing lots of manipulations.

### DIFF
--- a/agave_app/AppearanceSettingsWidget.cpp
+++ b/agave_app/AppearanceSettingsWidget.cpp
@@ -413,6 +413,28 @@ QAppearanceSettingsWidget::createClipPlaneSection(QAction* pToggleRotateAction, 
 
   sectionLayout->addRow("Hide", m_hideUserClipPlane);
 
+  // Add the Reset button
+  m_clipPlaneResetButton = new QPushButton("Reset");
+  m_clipPlaneResetButton->setStatusTip(tr("Reset clip plane to 0,0,0,0"));
+  m_clipPlaneResetButton->setToolTip(tr("Reset clip plane to 0,0,0,0"));
+  QObject::connect(m_clipPlaneResetButton, &QPushButton::clicked, [this]() {
+    if (!this->m_scene || !this->m_scene->m_clipPlane) {
+      return;
+    }
+    glm::vec3 c = this->m_scene->m_boundingBox.GetCenter();
+    this->m_scene->m_clipPlane->resetTo(c);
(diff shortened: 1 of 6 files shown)


agave_app/GLView3D.cpp
```diff
@@ -311,8 +311,9 @@ GLView3D::OnSelectionChanged(SceneObject* so)
   // has the effect of re-creating the manipulator tool,
   // which will effectively call origins.update to get the new
   // selection into the tool
+  MANIPULATOR_MODE mode = m_manipulatorMode;
   setManipulatorMode(MANIPULATOR_MODE::NONE);
-  setManipulatorMode(m_manipulatorMode);
+  setManipulatorMode(mode);
```
toloudis, line 316: the changes in this file are actually fixing a systematic bug, where I wasn't actually saving the previous manipulation mode.

meganrm (approved):

agave_app/AppearanceSettingsWidget.cpp
```diff
@@ -413,6 +413,28 @@ QAppearanceSettingsWidget::createClipPlaneSection(QAction* pToggleRotateAction,
...
+    if (!this->m_scene || !this->m_scene->m_clipPlane) {
+      return;
+    }
+    glm::vec3 c = this->m_scene->m_boundingBox.GetCenter();
+    this->m_scene->m_clipPlane->resetTo(c);
+    m_qrendersettings->renderSettings()->m_DirtyFlags.SetFlag(RoiDirty);
+    // re-select to cause Origins.update to reset the translate or rotate tools
+    if (this->m_scene->m_selection == this->m_scene->m_clipPlane.get()) {
+      emit this->m_qrendersettings->Selected(this->m_scene->m_clipPlane.get());
+    }
```
ShrimpCryptid, line 430: Oh, will it not reset it otherwise?

ShrimpCryptid (approved):

agave_app/AppearanceSettingsWidget.cpp
toloudis: the plane gets reset but if this happens during translate or rotate mode, the manipulator gizmo doesn't automatically know to move with the plane.  This function call causes that to happen.  Possibly, the future ui refactor may make this more automatic.

PR #257, opened 2025-04-17
Fix/clip plane hide

toloudis:
"Hide" checkbox for clip plane was not working properly.

Correct behavior after this bugfix:
Check the main clip plane checkbox to enable/disable clipping. When unchecked, clip plane will never show.

When checked (clipping is enabled):
  1. If user clicks Rotate or Translate, when in rotate/translate mode, we will always show the clip plane. This is so that users have a guide while rotating/translating.
  2. If "Hide" is checked, then when clip plane is enabled but not in translate/rotate mode, the clip plane will not be shown.

This is all achieved by:
1. always allow the clip plane's ManipulatorTool to have a chance to draw (if the clip plane is enabled)
2. it will only draw if the clip plane is selected or if hide is unchecked.
3. we also have to be careful about what is selected when rotate/translate is clicked, since there is also a rotate/translate button for area lights

### DIFF
--- a/renderlib/ViewerWindow.cpp
+++ b/renderlib/ViewerWindow.cpp
@@ -134,9 +134,9 @@ ViewerWindow::update(const SceneView::Viewport& viewport, const Clock& clock, Ge
   // [...]
 
   // TODO FIXME
-  // gross ugly disgusting hack
   // We need a mechanism to add a tool just from a scene object that was added to the scene.
   // For now, we just add the tool for special scene objects(like clip plane) right here on the fly.
+  // Instead of adding to this temporary vector, we should add to the sceneView.scene->m_tools
   std::vector<ManipulationTool*> sceneTools;
   if (sceneView.scene) {
     if (sceneView.scene->m_clipPlane) {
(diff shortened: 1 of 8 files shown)


renderlib/ViewerWindow.cpp
```diff
@@ -149,6 +169,9 @@ ViewerWindow::update(const SceneView::Viewport& viewport, const Clock& clock, Ge
 
   if (pickedAnything) {
     int selectionCode = gesture.graphics.getCurrentSelectionCode();
+    for (ManipulationTool* tool : sceneTools) {
+      tool->setActiveCode(selectionCode);
+    }
     forEachTool([&](ManipulationTool* tool) { tool->setActiveCode(selectionCode); });
```
ShrimpCryptid, line 175: Is this line redundant now?

toloudis: No, the code is now walking two separate lists.  the forEachTool still loops over the member variable that holds a list of tools.

ShrimpCryptid (approved):

frasercl (approved):
LGTM

PR #262, opened 2025-05-14
Feature/python self launch

### LINKED ISSUE #260: python scripts can self-launch agave server mode
## Use Case

Python scripts would be more self-contained and it would skip steps for a user if the script launched its own agave subprocess.

## Solution

Modify the pythonclient to allow users to provide an agave path, or try to find an agave install, in addition to just connecting to a url.  Handle errors.

toloudis:
Time to review:  20 min? 
Line count is large but most of it is in newly added test files. 

Changes: 

Modifications to agave_pyclient so that a Python script exported from AGAVE can run all by itself, without users having to explicitly launch AGAVE separately in server mode.

Before, users had to do something like:
```
agave --server
python -m myscript.py # connects to the above server instance using an assumed localhost port 
```
And now, it's just:
```
python myscript.py # launches its own AGAVE instance and closes it when script is done
```

AGAVE's python export now inserts the executable path so that the script can self-launch AGAVE.
Users are still required to install AGAVE separately from the agave_pyclient package.

The python package is also changed to use pyproject.toml and ruff, where before it used setup.py and flake8/black.

The most interesting changes are in `agave.py` and `find_agave.py`.


agave_app/ViewerState.cpp
```diff
@@ -91,18 +92,34 @@ stateToPythonScript(const Serialize::ViewerState& s)
...
+    if (c == '\\' || c == '"') {
+      escapedAppPath.push_back('\\');
+    }
+    escapedAppPath.push_back(c);
+  }
+  // TODO give user option to save a self-contained script or require agave server
+  // For now, most users will probably not mind that we are putting the executable
+  // path in the Python script.
+  // Pass auto_launch=False or omit agave_path to customize this behavior.
+  ss << "r = agave.AgaveRenderer(agave_path=\"" << escapedAppPath << "\", mode=\"" << mode << "\")" << std::endl;
```
toloudis, line 120: this is where we pass the current executable path through to the python code

agave_pyclient/agave_pyclient/agave.py
```diff
@@ -158,30 +93,130 @@ class AgaveRenderer:
...
+    def __init__(
+        self,
+        url: str = "ws://localhost:1235/",
+        mode: str = "pathtrace",
+        agave_path: Optional[str] = None,
+        auto_launch: bool = True,
+        launch_retries: int = 10,
+        launch_retry_delay: float = 1.0,
+    ) -> None:
+        # Connection precedence (highest to lowest):
```
toloudis: this comment describes the most important changes of this PR

agave_pyclient/agave_pyclient/agave_client.py
```diff
@@ -0,0 +1,90 @@
...
+import queue
+
+from ws4py.client.threadedclient import WebSocketClient
+
+__all__ = ["AgaveClient"]
+
+
+# assumptions: every commandbuffer send should result in one image.
+# also, they arrive in the order the buffers were sent.
+class AgaveClient(WebSocketClient):
```
toloudis, line 19: this class was moved from agave.py with no modifications

meganrm (approved):

.github/workflows/build_and_test.yml
```diff
@@ -229,15 +229,20 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
```
ShrimpCryptid, line 230: Outside scope but consider pinning to specific commit hash for security

agave_app/ViewerState.cpp
```diff
@@ -142,6 +159,7 @@ stateToPythonScript(const Serialize::ViewerState& s)
           .toPythonString()
      << std::endl;
   // TODO use value from viewport or render window capture settings?
+  // if user never went into capture, then this will be 32 the default, and will not re-create the viewport render.
```
ShrimpCryptid:
```suggestion
  // if user never went into capture, then this will be 32 by default, and will not re-create the viewport render.
```

agave_pyclient/agave_pyclient/bin/example2.py
```diff
@@ -0,0 +1,122 @@
...
+        r.skylight_top_color(1, 1, 1)
+        r.skylight_middle_color(1, 1, 1)
+        r.skylight_bottom_color(1, 1, 1)
+        r.light_pos(0, 1.5, 5.54884, 0.952752)
+        r.light_color(0, 10, 10, 10)
+        r.light_size(0, 0.15, 0.15)
+        for i in range(1, 65):
+            r.set_clip_region(0, 1, 0, 1, 0, i / 65)
+            r.session(f"1_{i}.zarr.png")
+            r.redraw()
```
ShrimpCryptid, line 118: Question for my own curiosity, I can tell this is expanding a clip region to show a whole volume. Does `r.session()` mark a path to save out to, and `r.redraw` initiates the render for that frame?

agave_pyclient/pyproject.toml
```diff
@@ -0,0 +1,73 @@
+[build-system]
+requires = ["setuptools>=64", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "agave_pyclient"
+version = "1.9.0"
+description = "A Python client for the Agave 3d volume renderer"
```
ShrimpCryptid: AGAVE since it's an acronym?

agave_pyclient/agave_pyclient/tests/test_commandbuffer.py
```diff
@@ -1,2 +1,331 @@
...
+def test_compute_size_scalar_args():
+    cb = CommandBuffer()
+    cb.add_command("EYE", 1.0, 2.0, 3.0)
+    assert cb.compute_size() == 4 + 3 * 4
+
+
+def test_compute_size_string_arg():
+    cb = CommandBuffer()
+    cb.add_command("SESSION", "hi")
+    assert cb.compute_size() == 4 + 4 + 2
```
ShrimpCryptid, line 60: Do you need to test for non-ASCII characters?

agave_pyclient/agave_pyclient/agave.py
```diff
@@ -158,30 +93,137 @@ class AgaveRenderer:
...
+
+       a. ``agave_path`` if explicitly provided, else
+       b. :func:`guess_agave_path` (PATH, then standard install
+          locations for the current OS).
+
+       The executable is launched with ``--server --port=<port>``
+       where ``<port>`` is parsed from ``url``, then the connection
+       is retried up to ``launch_retries`` times spaced
+       ``launch_retry_delay`` seconds apart.
     """
```
ShrimpCryptid, line 149: Big fan of this documentation

agave_pyclient/agave_pyclient/agave_client.py
```diff
@@ -0,0 +1,90 @@
...
+        super(AgaveClient, self).__init__(*args, **kwargs)
+        self.onOpened = None
+        self.onClose = None
+        self.messages = queue.Queue()
+
+    def load_image(self, image_path, onLoaded=None):
+        self.get_info(image_path, callback=onLoaded)
+
+    def opened(self):
+        print("opened up")
```
ShrimpCryptid, line 30: Are these debug messages still necessary?

ShrimpCryptid (approved):
Skimmed the unit tests, left some questions/comments!

agave_pyclient/agave_pyclient/bin/example2.py
toloudis: Yes, session basically just stores the path, and redraw tells agave to generate an image and return it via websockets.  On the python side, redraw also waits for that image to be returned.

agave_pyclient/pyproject.toml
toloudis: good idea!

agave_pyclient/agave_pyclient/tests/test_commandbuffer.py
toloudis: I think that opens a can of worms that might cause a lot of other changes.  I will create a new issue ticket for this.

agave_pyclient/agave_pyclient/bin/example2.py
toloudis: The reason why there is a special "session" command is slightly legacy.  There is a web implementation that actually streams iterative frames back to the web canvas, and so in that mode, one session can receive many many images.  The session id can also be correlated with a connection id, and in that way could be used to help generate multiple independent renderings (different images, even) from the same agave server.

agave_pyclient/agave_pyclient/tests/test_commandbuffer.py
toloudis: #352

agave_pyclient/agave_pyclient/agave_client.py
toloudis: They could use some cleanup.  Probably should be changed to use the logger.  They do give useful status when running the script, which is reassuring to see. I'll make this a ticket for another pr.

toloudis: #353

PR #269, opened 2025-06-11
fix the frame view function

toloudis:
Time to review: 5 min or less.
 
 Fix the "frame view" functionality.
 I believe a previous change in 1.8 development caused the sense of `Plane.d` to be negated but there was still this function that was explicitly using the d value.   

This PR also fixes an issue where you can click "Frame View" with nothing loaded and it crashed the app.

### DIFF
--- a/renderlib/graphics/glsl/GLThickLines.cpp
+++ b/renderlib/graphics/glsl/GLThickLines.cpp
@@ -1,12 +1,10 @@
 #include "GLThickLines.h"
 
+#include "glm.h"
 #include "shaders.h"
 
 #include <vector>
 #include <string>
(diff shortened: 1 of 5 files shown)


renderlib/graphics/glsl/GLThickLines.cpp
```diff
@@ -1,12 +1,10 @@
 #include "GLThickLines.h"
 
+#include "glm.h"
```
toloudis, line 3: these changes are just a cleanup thing, where my "glm.h" itself already includes several of the "<glm/glm....hpp>" includes.

meganrm (approved):

ascibisz (approved):

PR #279, opened 2025-07-18
fix docs builds

toloudis:
Time to review:  fairly quick.  

Docs builds broke due to tech debt with the action I was using to run Sphinx document generator.
This change switches to a more manual docs generation process.

Verified by: run build, confirm that docs build artifacts look like the old docs.
The build actions run two different jobs for docs:  `docs` and `docs_pythonclient`

Please disregard broken windows c++ build, which is due to a completely different dependency situation.

Reviewers: There may be extra dependencies thrown in; I didn't attempt to minimize the set of things installed to run this.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -10,14 +10,29 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: "recursive"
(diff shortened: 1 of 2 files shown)


interim17:
What degree of local testing would you like to confirm these changes?

PR #280, opened 2025-07-18
Optimize for less than 4 channels

toloudis:
The first of the two changes below can be a big optimization for loading single-channel volumes.  It basically means you can load 4x the amount of volume data with the same performance now, if it's single channel. 

Two key changes here: 
1. Previously AGAVE was always allocating a buffer that holds 4 channels of volume data -- even if you only load a single channel volume.  In this change, I modify gpu memory usage so that if less than 4 channels are really being loaded by AGAVE, then allocate a gpu buffer that holds less than 4 channels of volume data.  The code change also updates the calculated memory estimate that AGAVE provides to the user at load time.

2. The ui 3D manipulator tool code was creating and destroying some small buffers (a ScopedGLVertexBuffer and a ScopedGLTextureBuffer) on every single redraw of the ui tools.   I make a change here to reuse the buffers and only resize/reallocate if the number of elements to draw is increasing.

Bonus changes that snuck in:
* handle the case where a file is attempted to load with zero scenes present.  This can happen with some tiff files and wasn't handling the error properly.
* CI build: update windows sdk version
* tiny code cleanups

### DIFF
--- a/renderlib/io/CMakeLists.txt
+++ b/renderlib/io/CMakeLists.txt
@@ -34,8 +34,10 @@ add_subdirectory(libCZI)
 set(TENSORSTORE_USE_SYSTEM_TIFF ON)
 FetchContent_Declare(
 	tensorstore
-        URL "https://github.com/google/tensorstore/archive/refs/tags/v0.1.71.tar.gz"
-        URL_HASH SHA256=997ae179a3e6203c745afe9dfd5a1594ce5cf1738ebfafd3031f5f23211929f2
+        # URL "https://github.com/google/tensorstore/archive/refs/tags/v0.1.71.tar.gz"
+        # URL_HASH SHA256=997ae179a3e6203c745afe9dfd5a1594ce5cf1738ebfafd3031f5f23211929f2
+        URL "https://github.com/google/tensorstore/tarball/9e1feb7215ee876d3b62b8d84c03c9c2a1eec0c2"
+        URL_HASH SHA256=baeb9e1eafe1d8244d37c614711a8125467404801274e35979ba2896c8abdf29
 )
 
 # Additional FetchContent_Declare calls as needed...


renderlib/graphics/GestureGraphicsGL.cpp
```diff
@@ -347,12 +401,17 @@ GestureRendererGL::draw(SceneView& sceneView, SelectionBuffer* selection, Gestur
   // Draw UI and viewport manipulators
   {
     // TODO are we really creating, uploading, and destroying the vertex buffer every frame?
```
ShrimpCryptid, line 403: Remove TODO?

renderlib/graphics/GestureGraphicsGL.cpp
```diff
@@ -347,12 +401,17 @@ GestureRendererGL::draw(SceneView& sceneView, SelectionBuffer* selection, Gestur
...
-    ScopedGlVertexBuffer vertex_buffer(graphics.verts.data(),
-                                       graphics.verts.size() * sizeof(Gesture::Graphics::VertsCode));
+    // ScopedGlVertexBuffer vertex_buffer(graphics.verts.data(),
+    //                                    graphics.verts.size() * sizeof(Gesture::Graphics::VertsCode));
```
ShrimpCryptid, line 405: Remove?

renderlib/graphics/ImageXyzcGpu.cpp
```diff
@@ -87,7 +95,12 @@ ImageGpu::createVolumeTextureFusedRGBA8(ImageXYZC* img)
 void
 ImageGpu::createVolumeTexture4x16(ImageXYZC* img)
 {
-  m_gpuBytes += (16 * 4) / 8 * img->sizeX() * img->sizeY() * img->sizeZ();
+  int N = 4;
+  if (img->sizeC() < 4) {
+    N = img->sizeC();
+  }
```
ShrimpCryptid, line 101: Equivalent to doing `N = std::min(4, img->sizeC())`, right?
(I guess not worth replacing since you'd need to include `<algorithm>`)

ShrimpCryptid (approved):
LGTM! Left a few minor questions/nits

agave_app/loadDialog.cpp
```diff
@@ -227,7 +227,8 @@ LoadDialog::updateMemoryEstimate()
   spec.maxy = m_roiY->maxValue();
   spec.minz = m_roiZ->minValue();
   spec.maxz = m_roiZ->maxValue();
-  size_t mem = spec.getMemoryEstimate();
+  spec.channels = getCheckedChannels();
```
interim17, line 230: So this is used later but isn't necessary for memory estimate below?

interim17 (approved):
Left one comment but it's definitely non-blocking as it's largely over my head.

renderlib/graphics/Framebuffer.cpp
```diff
@@ -70,6 +70,9 @@ Framebuffer::resize(uint32_t w, uint32_t h)
 void
 Framebuffer::destroyFb()
 {
+  glBindFramebuffer(GL_FRAMEBUFFER, 0);
+  check_gl("bind 0 framebuffer");
```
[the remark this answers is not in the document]
toloudis: yes, check_gl calls into OpenGL's error mechanism and then if an error code is there, it elevates it to a logged console error message but nothing more.  It's a cheap call when no error happened.  Most useful for weird graphics state errors that a user wouldn't understand but I would.

renderlib/graphics/GestureGraphicsGL.cpp
```diff
@@ -572,3 +633,20 @@ GestureRendererGL::pick(SelectionBuffer& selection,
...
   return entry != Gesture::Graphics::k_noSelectionCode;
 }
+
+GestureRendererGL::GestureRendererGL() {}
+
+GestureRendererGL::~GestureRendererGL()
+{
+  // Destroy OpenGL resources
+  vertex_buffer.~ScopedGlVertexBuffer();
+  texture_buffer.~ScopedGlTextureBuffer();
```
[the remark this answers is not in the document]
toloudis: you are correct to flag this as weird. We don't normally call destructors explicitly like this. I had this in for a reason at first but I'm going to look at the code again and try to remember why.  It really shouldn't be here ideally.

Normally a class destructor is where all its member variables go out of scope forever, and their destructors are automatically called, so one would normally never do it explicitly like this.

agave_app/loadDialog.cpp
toloudis: spec.channels is used inside of getMemoryEstimate

Issue #283, opened 2025-08-15
bounding box lines are too thin


toloudis:
## Use Case

The bounding box is very thin at one pixel, especially when rendering at fairly high resolutions.  Would like thicker lines or some user control over this.

## Solution

Can add a user setting for a line thickness factor, but also make the bounding box use the newly added thick-lines  rendering.

PR #285, opened 2025-09-17
Feature/bbox thick lines

toloudis:
This is part 1 of two change sets to address #283 .
* Add a BoundingBoxTool that can draw box lines as "thick" lines.  
* Currently this is implemented only for the front-facing box edges and tickmarks (which always appear "on top" of the rest of the scene).  The code works on a per-edge basis.  The back-facing edges are still single-pixel thin GL_LINES and are still rendered inside the volume renderer because of how they have to blend with the rendered volume.  Part 2 will address that.
* Do some refactoring for calculating tick-marks so that both bounding box rendering methods can draw them consistently.  (See GetEdgeTickMarkVertices in BoundingBox.cpp)

### DIFF
--- a/renderlib/BoundingBox.h
+++ b/renderlib/BoundingBox.h
@@ -266,7 +266,7 @@ public:
 	bool IntersectP(const CRay& ray, float* hitt0 = NULL, float* hitt1 = NULL)
 	{
 		float t0 = ray.m_MinT, t1 = ray.m_MaxT;
-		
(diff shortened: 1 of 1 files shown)


toloudis:
> overall seems good, but you have build errors

Sadly, the build error has nothing to do with the changes here.  See commentary in #290 .  Long story short: Intel macbooks are getting old.

PR #297, opened 2025-10-23
Feature/tfeditor improved

toloudis:
Transfer Function editor now gets some major improvement.

1. The appearance is clearer now.  Background is light, histogram is blue, and transfer function is black.
2. The graph can be zoomed with scrollwheel and panned by dragging left/right.  Double click will re-zoom to fit (resetting back to default)
3. Control points can now be dragged horizontally in Min/max, Window/level, and Percentile modes. Sliders should auto update to match.
4. x axis of graph gets tickmarks with true raw intensity values

**Do not review the files qcustomplot.cpp and qcustomplot.h; it is third-party code.**

The main code changes are in `gradients.cpp` and `gradients.h`.  
Hoverpoints is completely removed.
So is ShadeWidget.
All functionality of hoverpoints and shadewidget is now in the GradientEditor and QCustomPlot.

TODO:
The plotting component I am using to do the new transfer function is GPL licensed, which implies that AGAVE must now adopt GPL.  Future PR to come...

Before:

Screenshot 2025-10-26 at 8 19 37 PM

After:

Screenshot 2025-10-26 at 8 19 05 PM

### DIFF
--- a/agave_app/tfeditor/gradients.cpp
+++ b/agave_app/tfeditor/gradients.cpp
@@ -71,9 +71,17 @@ GradientEditor::GradientEditor(const Histogram& histogram, QWidget* parent)
   myBars->setWidth(binSize);
   QVector<double> keyData;
   QVector<double> valueData;
+  static constexpr double MIN_BAR_HEIGHT = 0.01; // Minimum height for nonzero bins (0.1% of max)
   for (size_t i = 0; i < histogram._bins.size(); ++i) {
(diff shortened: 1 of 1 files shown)


agave_app/tfeditor/gradients.h
toloudis: this file has changed so significantly by now, that the old copyright notice in here is no longer relevant.

frasercl (approved):

PR #308, opened 2026-01-18
Fix/threshold time series

### LINKED ISSUE #310: outlier pixels mess up whole frames during time series rendering
## Description

time series with one bad extremely high intensity pixel can blow out the transfer function for a whole frame.  This seems to be because of histogram recalculation and re-binning resulting in huge bin widths.

## Expected Behavior

expect consistent intensity across frames in a time series even if there are occasional outlier pixels.

toloudis:
Resolves #310 and #309 : rendering bug when there are outlier pixel intensities, and histogram not updating in the channel settings display.

Key bug fixes / improvements around histograms and thresholding.

1. displayed per-channel histograms were not updating when a Time change happens in a time series image.  Add plumbing to allow the gradient widgets to update the histogram graph.
2. outlier pixel intensities could create unusably large bin widths in which all the "signal" pixels ended up in a very few bins.  Introduce percentile outlier detection to improve this.
3. The rendered lookup-table transfer function was too sensitive to changes in data range related to item 2 above.  I changed it to use a piecewise-linear function lookup with control points. (Supports a max of 16 control points).  
3a. For the time being, the old transfer function (which was a lookup table stored in a texture) is still in place, but deprecated and can be removed from the code later.
4. There are some situations where control points seemed to end up in unsorted order.  Add sorting to a few key places.

### DIFF
--- a/renderlib/graphics/glsl/shadersrc/pathTraceVolume.frag
+++ b/renderlib/graphics/glsl/shadersrc/pathTraceVolume.frag
@@ -71,7 +71,7 @@ uniform float gGradientFactor;
 uniform float uShowLights;
 
 // per channel
-uniform sampler2D g_lutTexture[4];
+uniform sampler2D g_lutTexture[4]; // DEPRECATED to be removed in future stable release
 uniform sampler2DArray g_colormapTexture;
 
 // only used for lut Texture lookups
@@ -87,7 +87,9 @@ uniform vec3 g_diffuse[4];
 uniform vec3 g_specular[4];
 uniform float g_roughness[4];
 
+// this const must be kept in sync with GLPTVolumeShader.cpp
 const uint MAX_NO_TF_NODES = 16u;
+
 // Each node is a vec2: x = intensity, y = value
 // composing a piecewise linear transfer function.
 // This assumes that the x values are sorted in ascending order!!!
@@ -132,6 +134,7 @@ evalTf(in uint channel, in float intensity)
   return 0.0f;
 }
 
+// DEPRECATED to be removed in future stable release
 float
 evalTfLut(in uint channel, in float intensity)
 {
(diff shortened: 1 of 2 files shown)


agave_app/AppearanceSettingsWidget.cpp
```diff
@@ -1271,3 +1274,20 @@ QAppearanceSettingsWidget::onNewImage(Scene* scene)
...
+  // update all gradient widgets with new histogram for time point
+  if (!m_scene || !m_scene->m_volume) {
+    return;
+  }
+
+  for (size_t i = 0; i < m_gradientWidgets.size(); ++i) {
+    GradientWidget* gradientWidget = m_gradientWidgets[i];
+    if (gradientWidget && i < m_scene->m_volume->sizeC()) {
+      // Update histogram from current channel
+      gradientWidget->setHistogram(m_scene->m_volume->channel(i)->m_histogram);
```
toloudis: this line actually causes the graph to be updated

renderlib/graphics/glsl/shadersrc/pathTraceVolume.frag
```diff
@@ -92,6 +103,67 @@ uniform sampler2D tPreviousTexture;
 
 uniform vec4 g_clipPlane;
 
+float
+evalTf(in uint channel, in float intensity)
```
toloudis: this is the function that evaluates a piecewise linear transfer function defined as sets of control points

renderlib/Histogram.cpp
```diff
@@ -34,48 +31,121 @@ Histogram::getBinOfIntensity(uint16_t intensity) const
...
+        _dataMax = i;
+        break;
+      }
+      // Prevent underflow on unsigned decrement
+      if (i == 0)
+        break;
+    }
+
+    // Ensure we have a valid range
+    if (_dataMin >= _dataMax) {
```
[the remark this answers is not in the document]
toloudis: I think after outlier filtering it is possible that the datamin and datamax are equal. i guess impossible for max<min.

ShrimpCryptid (approved):
LGTM!

PR #351, opened 2026-05-18
Fix/tfeditor time change

toloudis:
Bug fixes and cleanup for the transfer function editor:

1. when new data was loaded at new time points, some sliders were not updating properly.  That issue is fixed with a slider update.
2. it was possible to move the min handle and see the max handle also shift unpredictably.  This is fixed by rewriting some of the code that handles dragging the control points: making sure that only one point is actually being updated.
3. add to the documentation about what happens to the transfer functions when time changes and new time series volume data arrives.
4. minor update to agents.md file

### DIFF
--- a/agave_app/tfeditor/gradients.cpp
+++ b/agave_app/tfeditor/gradients.cpp
@@ -1429,13 +1429,10 @@ GradientWidget::onSetIsovalue(float isovalue, float width)
 void
 GradientWidget::onInteractivePointsChanged(float minIntensity, float maxIntensity, int draggedIndex)
 {
-  // Only update the side that was actually dragged. Reading back the un-dragged side's
-  // editor key and writing it through to the slider/data causes accumulating drift,
-  // because the editor key is a float reconstruction (dataMin + relative*range) of the
-  // stored uint16/percentile and the inverse mapping is lossy:
-  //   - MINMAX:     uint16 -> float -> uint16  (off-by-one truncation drift)
-  //   - PERCENTILE: pct -> intensity -> bin -> pct  (drift up to one histogram bin width,
-  //                 which is far more than 1 intensity unit, and persists across drags).
+  // Only update the side that was actually dragged.
+  // This is to avoid doing unnecessary updating of values that shouldn't be changing anyway.
+  // The math to update the values could actually cause rounding errors which cause drift.
+
   // draggedIndex: 1 = low/min threshold, 2 = high/max threshold.
   const bool draggingLow = (draggedIndex == 1);
   const bool draggingHigh = (draggedIndex == 2);
--- a/agave_app/tfeditor/gradients.cpp
+++ b/agave_app/tfeditor/gradients.cpp
@@ -1328,24 +1328,15 @@ GradientWidget::onGradientStopsChanged(const QGradientStops& stops)
               m_gradientData->m_customControlPoints.end(),
               controlpoint_x_less_than);
     emit gradientStopsChanged(stops);
-  } else if (m_gradientData->m_activeMode == GradientEditMode::WINDOW_LEVEL) {
-    // extract window and level from the stops - use second and third points (threshold points)
-    std::vector<LutControlPoint> points = gradientStopsToVector(const_cast<QGradientStops&>(stops));
-    if (points.size() < 4) {
-      return;
-    }
-    std::sort(points.begin(), points.end(), controlpoint_x_less_than);
-    float low = points[1].first;  // Second point (low threshold)
-    float high = points[2].first; // Third point (high threshold)
-    float window = high - low;
-    float level = (high + low) * 0.5f;
-    m_gradientData->m_window = window;
-    m_gradientData->m_level = level;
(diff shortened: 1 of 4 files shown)


agave_app/tfeditor/gradients.cpp
```diff
@@ -1444,49 +1427,64 @@ GradientWidget::onSetIsovalue(float isovalue, float width)
 }
 
 void
-GradientWidget::onInteractivePointsChanged(float minIntensity, float maxIntensity)
+GradientWidget::onInteractivePointsChanged(float minIntensity, float maxIntensity, int draggedIndex)
```
toloudis, line 1430: this is the function that was previously updating both the min and max regardless of which point was dragged.

agave_app/tfeditor/gradients.cpp
```diff
@@ -1292,24 +1328,15 @@ GradientWidget::onGradientStopsChanged(const QGradientStops& stops)
...
-
+  } else if (m_gradientData->m_activeMode == GradientEditMode::WINDOW_LEVEL ||
+             m_gradientData->m_activeMode == GradientEditMode::PERCENTILE ||
+             m_gradientData->m_activeMode == GradientEditMode::MINMAX) {
+    // Threshold modes: do NOT update m_gradientData or sliders here. That is done by
+    // onInteractivePointsChanged with knowledge of which handle was dragged, so we don't
+    // corrupt the un-dragged side via a lossy float / percentile round-trip and don't
+    // re-trigger set_shade_points mid-drag (which would visually move the un-dragged
+    // handle). Just forward the stops so the renderer's LUT updates live during the drag.
+    emit gradientStopsChanged(stops);
```
toloudis, line 1339: this is a key part of the bugfix, that prevents unpredictable movement of the control points due to the roundtrip described in the comment.

agave_app/tfeditor/gradients.cpp
```diff
@@ -1075,8 +1076,43 @@ GradientWidget::GradientWidget(const Histogram& histogram, GradientData* dataObj
...
+
+    windowSlider->setValue(m_gradientData->m_window);
+    levelSlider->setValue(m_gradientData->m_level);
+    isovalueSlider->setValue(m_gradientData->m_isovalue);
+    isorangeSlider->setValue(m_gradientData->m_isorange);
+    pctLowSlider->setValue(m_gradientData->m_pctLow);
+    pctHighSlider->setValue(m_gradientData->m_pctHigh);
+  }
+
+  forceDataUpdate();
```
toloudis, line 1115: the above code is the fix for time series data

agave_app/tfeditor/gradients.cpp
```diff
@@ -1444,49 +1427,64 @@ GradientWidget::onSetIsovalue(float isovalue, float width)
...
-      minu16Slider->blockSignals(true);
-      minu16Slider->setValue(minu16);
-      minu16Slider->blockSignals(false);
+  if (m_gradientData->m_activeMode == GradientEditMode::MINMAX) {
+    // clamp the dragged value to the data range
+    float clamped = std::max(dataMin, std::min(draggingLow ? minIntensity : maxIntensity, dataMax));
+    uint16_t value = static_cast<uint16_t>(std::round(clamped));
+    // update the dragged side's value in m_gradientData.
+    if (draggingLow) {
+      value = std::min(value, m_gradientData->m_maxu16);
```
ShrimpCryptid, line 1452: Since min and max can swap, does the index of the handle that's getting dragged update accordingly?

agave_app/tfeditor/gradients.cpp
```diff
@@ -1444,49 +1427,64 @@ GradientWidget::onSetIsovalue(float isovalue, float width)
...
-    // Update the data
+    // window/level are coupled: derive the un-dragged endpoint from the stored window/level
+    // so it does NOT pick up float round-trip drift from the editor.
+    const float storedWindow = m_gradientData->m_window;
+    const float storedLevel = m_gradientData->m_level;
+    const float storedRelMin = storedLevel - 0.5f * storedWindow;
+    const float storedRelMax = storedLevel + 0.5f * storedWindow;
+
+    float relMin = storedRelMin;
+    float relMax = storedRelMax;
```
ShrimpCryptid: Why have `storedRelMax` and `storedRelMin`?

ShrimpCryptid (approved):

rugeli (approved):
lgtm!

agave_app/tfeditor/gradients.cpp
toloudis: I fixed this; they are clearly redundant

agave_app/tfeditor/gradients.cpp
toloudis: The min or max handles being dragged can not ever cross each other.  The swap is there for extra insurance but the mouse handler for dragging  (onPlotMouseMove) ensures this in minmax, percentile, and window/level modes, prior to when this code is called.

PR #367, opened 2026-05-22
Feature/release1.10cleanup

toloudis:
A handful of simple cleanup changes.  Most of the diffs are just changing the way tooltip strings are passed to controls, which covers a lot of lines.

1. Tooltips and statustips were not showing for UI labels.  They were showing up on the input controls themselves. The bulk of changes in here are to make sure tooltips and statustips show up when hovering over the labels as well as the input controls.
Screenshot 2026-05-22 at 11 00 36 AM
(Sorry, the mouse cursor did not show up in my screenshot but it's hovering over the "Renderer" label)

2. Lock Lights to Camera is now ON by default.
Screenshot 2026-05-22 at 10 59 32 AM

3. Fix misalignment of Transfer Function editor controls.  The sliders showing up below the graph editor buttons were too narrow and are now the same width as the button array.
Screenshot 2026-05-22 at 10 57 32 AM

4. Fix horizontal scrollbar.  When expanding a channel section, a horizontal scrollbar used to appear. This change stops that from happening.   A vertical scrollbar DOES appear (correctly indicating that some content is out of view), which actually was narrowing the panel and causing the horizontal scrollbar in the first place.
Screenshot 2026-05-22 at 10 58 30 AM

### DIFF
--- a/agave_app/tfeditor/gradients.cpp
+++ b/agave_app/tfeditor/gradients.cpp
@@ -892,6 +892,7 @@ GradientWidget::GradientWidget(const Histogram& histogram, GradientData* dataObj
   btnGroup->addButton(customButton, CUSTOM_BTNID);
   QHBoxLayout* hbox = new QHBoxLayout();
   hbox->setSpacing(0);
+  hbox->setContentsMargins(0, 0, 0, 0);
 
   int initialButtonId = WINDOW_LEVEL_BTNID;
   GradientEditMode m = m_gradientData->m_activeMode;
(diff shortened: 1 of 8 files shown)


agave_app/tfeditor/gradients.cpp
```diff
@@ -963,19 +969,15 @@ GradientWidget::GradientWidget(const Histogram& histogram, GradientData* dataObj
...
-  section0Layout->addRow("Min u16", minu16Slider);
+  Controls::addFormRow(section0Layout, "Min u16", minu16Slider, tr("Set minimum u16 value"), tr("Minimum u16 value"));
   maxu16Slider = new QIntSlider();
-  maxu16Slider->setStatusTip(tr("Maximum u16 value"));
-  maxu16Slider->setToolTip(tr("Set maximum u16 value"));
   maxu16Slider->setRange(m_histogram.getDataMin(), m_histogram.getDataMax());
   maxu16Slider->setSingleStep(1);
   maxu16Slider->setValue(m_gradientData->m_maxu16);
-  section0Layout->addRow("Max u16", maxu16Slider);
+  Controls::addFormRow(section0Layout, "Max u16", maxu16Slider, tr("Set maximum u16 value"), tr("Maximum u16 value"));
```
meganrm: are these use facing messages? if so "Set maximum u16 value" seems opaque

meganrm (approved):

agave_app/tfeditor/gradients.cpp
toloudis: yep, good point.  Will fix.

frasercl (approved):
Nothing looks out of place in a relatively cursory scan

PR #372, opened 2026-06-07
[volumecache 02/12] Route volume loads through memory cache

toloudis:
Stacked draft PR 2 of 12 for image caching in AGAVE.

This PR wires volume loading call sites through the new memory cache and clears cached entries when the active load context changes.

### DIFF
--- a/renderlib/io/FileReader.cpp
+++ b/renderlib/io/FileReader.cpp
@@ -7,12 +7,10 @@
 #include "FileReaderZarr.h"
 #include "ImageXYZC.h"
 #include "Logging.h"
+#include "CacheManager.h"
 
 #include <chrono>
 #include <filesystem>
-#include <map>
-
-std::map<std::string, std::shared_ptr<ImageXYZC>> FileReader::sPreloadedImageCache;
 
 // return file extension as lowercase
 std::string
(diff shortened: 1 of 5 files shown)


renderlib/io/FileReader.cpp
```diff
@@ -66,30 +64,34 @@ FileReader::getReader(const std::string& filepath, bool isImageSequence)
...
 
 std::shared_ptr<ImageXYZC>
-FileReader::loadAndCache(const LoadSpec& loadSpec)
+FileReader::loadAndCache(const LoadSpec& loadSpec, std::shared_ptr<IFileReader> reader)
 {
-  // check cache first of all.
-  auto cached = sPreloadedImageCache.find(loadSpec.filepath);
-  if (cached != sPreloadedImageCache.end()) {
-    return cached->second;
+  auto cached = CacheManager::instance().findImage(loadSpec);
```
ShrimpCryptid, line 69: Just a design question, are there other patterns you considered besides Singleton for the CacheManager? I remember hearing people say it's an anti-pattern and can make testing harder.

ShrimpCryptid (approved):
Left one question, but LGTM!

renderlib/io/FileReader.cpp
toloudis: Yeah.  This is a good question.  In this app there is indeed one and only one cache.  For unit testing, it is still possible to instantiate the cache within the unit test suite, and direct its storage location to some temp testing folder.   But it's certainly also possible to make the cache a member variable of some object that basically has application lifetime,  or instantiate it in main() and pass it down to the file io library, etc etc.  That is something I can pursue but is probably suited to a later PR.  Currently there are no issues with it the way it is.  Also, "renderlib" itself is effectively a singleton as well, although it is accessed via all-static class methods as opposed to a "instance" member.  Different flavors of the same thing.  The other concern is multithreaded access, and the Cache here has mutexes and concurrency protections inside its implementation.

ascibisz (approved):

PR #374, opened 2026-06-07
[volumecache 04/12] Guard disk cache clearing with marker file

toloudis:
Stacked draft PR 4 of 12 for the refreshed volume cache work.

This adds global cache clearing functionality.  It also deposits and checks against special marker files to try to be safe about only deleting info that was cached by agave.

### DIFF
--- a/renderlib/CacheManager.cpp
+++ b/renderlib/CacheManager.cpp
@@ -14,6 +14,12 @@
 
 namespace {
 
+// Marker file written into any directory we manage as our own disk cache root.
+// clearDiskCache requires this file to be present before it will delete
+// anything, which protects against the user pointing the cache dir at a path
+// like "C:\" or "/home/me" and then clicking "Clear disk cache".
+constexpr const char* kCacheMarkerFilename = ".agave-cache-dir";
+
 inline void
 hashCombine(std::size_t& seed, std::size_t value)
 {
@@ -227,7 +233,13 @@ CacheManager::setConfig(const CacheConfig& config)
     m_entries.clear();
     m_lruKeys.clear();
     m_currentRamBytes = 0;
+    m_diskEntries.clear();
+    m_currentDiskBytes = 0;
     return;
+  }
+  if (!m_config.enableDisk || m_cacheDir.empty()) {
+    m_diskEntries.clear();
+    m_currentDiskBytes = 0;
   }
   evictIfNeededLocked(0);
 }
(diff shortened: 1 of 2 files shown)


renderlib/CacheManager.cpp
```diff
@@ -289,6 +301,89 @@ CacheManager::clearMemoryCache()
...
+  }
+
+  if (cacheDir.empty()) {
+    return;
+  }
+
+  // Remove the per-entry subdirectories we know about.
+  for (const auto& path : knownEntryPaths) {
+    std::error_code ec;
+    std::filesystem::remove_all(path, ec);
```
ShrimpCryptid, line 334: Should this also happen inside of the mutex lock? I'm thinking about this possibility:
1. Lock releases, with `file_a` marked for deletion (stored in knownEntryPaths)
2. Another process starts caching `file_a` before L332-334 executes
3. `file_a` is deleted, even though the cache thinks that it is stored, leading to errors when fetching the cached file.

renderlib/CacheManager.cpp
```diff
@@ -289,6 +301,89 @@ CacheManager::clearMemoryCache()
...
+  }
+
+  // Also remove any orphan per-entry subdirectories left behind by prior
+  // sessions or partial writes. We only touch subdirectories that contain a
+  // meta.json (i.e. look like cache entries we wrote) — anything else the user
+  // may have placed in the cache dir is preserved.
+  std::error_code dirEc;
```
ShrimpCryptid, line 341: Should the error codes here be logged somewhere?

renderlib/CacheManager.cpp
```diff
@@ -289,6 +301,89 @@ CacheManager::clearMemoryCache()
...
+  for (auto it = std::filesystem::directory_iterator(cacheDir, dirEc);
+       it != std::filesystem::directory_iterator() && !dirEc;
+       it.increment(dirEc)) {
+    if (!it->is_directory()) {
+      continue;
+    }
+    std::filesystem::path metaPath = it->path() / "meta.json";
```
ShrimpCryptid, line 348: Should `meta.json` be a constant?

renderlib/CacheManager.cpp
```diff
@@ -289,6 +301,89 @@ CacheManager::clearMemoryCache()
...
+    if (std::filesystem::exists(metaPath, existEc)) {
+      std::error_code rmEc;
+      std::filesystem::remove_all(it->path(), rmEc);
+    }
+  }
+}
+
+void
+CacheManager::writeCacheMarker(const std::string& path) const
+{
```
ShrimpCryptid, line 359: Nit: Consider specifying in-code that the path is a directory... either renaming `path -> directory` or the function to `writeCacheMarkerToDir`?

ShrimpCryptid (approved):
Left some non-blocking comments, but the change was understandable and easy to follow.

renderlib/CacheManager.cpp
toloudis: it currently only exists in this one place

interim17 (approved):

PR #376, opened 2026-06-07
[volumecache 06/12] Index and evict disk cache entries

toloudis:
Stacked draft PR 6 of 12 for the refreshed volume cache work.

This PR adds cold-start disk index loading, disk byte accounting, and LRU-style eviction when writes would exceed the configured disk cap.

The key thing here is that when we start up agave, we read the disk cache so that we have an in-memory index of it.

### DIFF
--- a/renderlib/CacheManager.cpp
+++ b/renderlib/CacheManager.cpp
@@ -263,20 +263,42 @@ CacheManager::getCacheDirectory() const
 void
 CacheManager::setConfig(const CacheConfig& config)
 {
+  CacheConfig configCopy;
+  std::string cacheDirCopy;
+  bool rebuildDiskIndex = false;
+  {
+    std::scoped_lock lock(m_mutex);
+    m_config = config;
+    if (!m_config.enabled) {
+      m_entries.clear();
+      m_lruKeys.clear();
+      m_currentRamBytes = 0;
+      m_diskEntries.clear();
+      m_currentDiskBytes = 0;
+      m_diskIndexRoot.clear();
+      return;
+    }
+
+    if (!m_config.enableDisk || m_cacheDir.empty()) {
+      m_diskEntries.clear();
+      m_currentDiskBytes = 0;
+      m_diskIndexRoot.clear();
+    } else if (m_diskIndexRoot != m_cacheDir) {
+      m_diskEntries.clear();
+      m_currentDiskBytes = 0;
+      m_diskIndexRoot = m_cacheDir;
+      rebuildDiskIndex = true;
+      configCopy = m_config;
+      cacheDirCopy = m_cacheDir;
+    }
+  }
+
+  if (rebuildDiskIndex) {
+    loadDiskIndex(configCopy, cacheDirCopy);
+    evictDiskIfNeeded(configCopy, 0);
+  }
+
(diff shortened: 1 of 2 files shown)


renderlib/CacheManager.cpp
ShrimpCryptid, line 833: Maybe would also want to flag for a disk index rebuild here

renderlib/CacheManager.h
```diff
@@ -140,5 +143,11 @@ class CacheManager
   std::unordered_map<std::string, DiskEntry> m_diskEntries;
   std::uint64_t m_currentDiskBytes = 0;
 
+  // Tracks the root the in-memory index was last built against (used to
+  // decide when a rebuild is needed).
+  // Set to m_cacheDir when the in-memory disk index is built, such as on startup.
+  // Clear this value to signal that the in-memory disk index is no longer valid.
+  std::string m_diskIndexRoot;
```
ShrimpCryptid, line 150: I feel slightly weird about `m_diskIndexRoot` being both the root of the disk index and a flag for when it needs to be rebuilt... If it's only ever set to `m_cacheDir` when active, could it just be a boolean flag for needsRebuild? Or is the concern that `m_cacheDir` might change from under it?

renderlib/CacheManager.cpp
```diff
@@ -265,20 +265,42 @@ CacheManager::getCacheDirectory() const
...
-    m_diskEntries.clear();
-    m_currentDiskBytes = 0;
+
+  if (rebuildDiskIndex) {
+    loadDiskIndex(configCopy, cacheDirCopy);
+    evictDiskIfNeeded(configCopy, 0);
   }
+
+  std::scoped_lock lock(m_mutex);
   evictIfNeededLocked(0);
```
ShrimpCryptid, line 304: Nit: Should this be wrapped in a scope, in case this function is ever extended?

renderlib/CacheManager.cpp
```diff
@@ -811,3 +837,133 @@ CacheManager::storeToDisk(const CacheKey& key,
...
+  if (cacheDir.empty() || !config.enableDisk) {
+    return;
+  }
+
+  std::filesystem::path root(cacheDir);
+  std::error_code ec;
+  std::filesystem::create_directories(root, ec);
+  if (!std::filesystem::exists(root)) {
+    return;
+  }
```
ShrimpCryptid, line 853: Should this throw an error or log a warning?

renderlib/CacheManager.cpp
```diff
@@ -811,3 +837,133 @@ CacheManager::storeToDisk(const CacheKey& key,
...
+    }
+    m_diskEntries.erase(it);
+
+    std::error_code ec;
+    std::filesystem::remove_all(path, ec);
+    if (ec) {
+      // On Windows this can fail if another thread (e.g. loadFromDisk)
+      // still holds tensorstore file handles into the same path. The
+      // bookkeeping has already been updated to reflect eviction; the
+      // stale files will be picked up by the next clearDiskCache.
```
ShrimpCryptid, line 952: When is `clearDiskCache` called? If it has to be manually triggered by the user, is it possible for the files to persist through multiple sessions with AGAVE and continually expand the cache?

PR #377, opened 2026-06-07
[volumecache 07/12] Add platform cache resource checks

toloudis:
Stacked draft PR 7 of 12 for the refreshed volume cache work.

This PR adds renderlib SystemInfo helpers for available memory and disk space checks across Windows, Linux, and macOS.

### DIFF
--- a/renderlib/SystemInfo.cpp
+++ b/renderlib/SystemInfo.cpp
@@ -1,0 +1,108 @@
+#include "SystemInfo.h"
+
+#include <filesystem>
+#include <fstream>
+#include <sstream>
+#include <string>
+
+#ifdef _WIN32
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#include <windows.h>
+#elif defined(__linux__)
+#include <sys/sysinfo.h>
+#else // macOS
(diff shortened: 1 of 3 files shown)


renderlib/SystemInfo.cpp
```diff
@@ -0,0 +1,107 @@
...
+      }
+    }
+  }
+  // Older kernels (<3.14) don't expose MemAvailable. Fall back to
+  // sysinfo, and at least add bufferram to freeram so we don't grossly
+  // under-report.
+  struct sysinfo info;
+  if (sysinfo(&info) == 0) {
+    std::uint64_t bytes = static_cast<std::uint64_t>(info.freeram) + static_cast<std::uint64_t>(info.bufferram);
+    return bytes * static_cast<std::uint64_t>(info.mem_unit);
```
ShrimpCryptid, line 60: Why multiply by `mem_unit` here? 

https://man7.org/linux/man-pages/man2/sysinfo.2.html

Linux manual says it's the memory unit size in bytes, but we're multiplying by bytes already...?

ShrimpCryptid (approved):

PR #383, opened 2026-06-09
give the statistics view a bit of margin

### LINKED ISSUE #178: Right padding missing
## Description

Right padding of the right panel is missing

## Expected Behavior

Needs padding

## Reproduction

Visible when looking at UI

## Environment

MacOS

Screenshot 2024-04-30 at 11 18 45 AM

toloudis:
quick to review

addresses #178 Right padding missing

adds a wrapper widget and gives it a right margin of 8px !

### DIFF
--- a/agave_app/StatisticsDockWidget.cpp
+++ b/agave_app/StatisticsDockWidget.cpp
@@ -2,10 +2,16 @@
 
 QStatisticsDockWidget::QStatisticsDockWidget(QWidget* pParent)
   : QDockWidget(pParent)
+  , m_MainWidget(this)
   , m_MainLayout()
-  , m_StatisticsWidget()
+  , m_StatisticsWidget(&m_MainWidget)
 {
   setWindowTitle("Statistics");
 
-  setWidget(&m_StatisticsWidget);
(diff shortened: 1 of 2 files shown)


rugeli (approved):
LGTM!

agave_app/StatisticsDockWidget.cpp
```diff
@@ -2,10 +2,16 @@
 
 QStatisticsDockWidget::QStatisticsDockWidget(QWidget* pParent)
   : QDockWidget(pParent)
+  , m_MainWidget(this)
```
rugeli, line 5: pure cpp/qt learning question, not blocking at all: why pass `this` as parent here, since `setWidget()`(L16) below should also make it a child?

toloudis: You're right to think this. But because this is a QDockWidget, setWidget does some Qt-specific layout things related to the docking behavior.  So it's not the same as just making it a parent.  The setWidget call is more important than the assignment of the parent in the m_MainWidget initializer.

ShrimpCryptid (approved):
Oops, thought I'd already left an approval on this!

PR #385, opened 2026-06-10
Fix gh actions nsis install

toloudis:
The previous github action I was using for this suddenly broke.  Other users have noticed this too:  https://github.com/repolevedavaj/install-nsis/issues/44 
That is the issue I am working around here.

NSIS is the windows installer builder software, which we need to create the windows installer for agave.  Previously was using a gh action to get it, but the choco method is more robust as it's a builtin package manager on the windows runner.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -168,9 +168,7 @@ jobs:
           choco install nasm
       - name: windows install nsis
         if: matrix.os == 'windows-latest'
-        uses: repolevedavaj/install-nsis@v1.2.0
-        with:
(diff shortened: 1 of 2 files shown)


meganrm (approved):

.github/workflows/build_and_test.yml
```diff
@@ -168,9 +168,7 @@ jobs:
           choco install nasm
```
ShrimpCryptid, line 168: Just curious, is there a reason why you don't need to include the `-y` flag on previous `choco` installs in this action?

ShrimpCryptid (approved):

PR #405, opened 2026-08-01
Feature/timeseries loading

toloudis:
logic for asyn loading-ahead of timesteps for time series data, and playback in-app


agave_app/Controls.cpp
```diff
@@ -453,21 +453,12 @@ QIntSlider::QIntSlider(QWidget* pParent /*= NULL*/)
...
-    // If the owner has disabled tracking, that's a good bet that the increments are
-    // expected to be "too long".
-    if (!this->m_slider.hasTracking()) {
-      this->m_spinner.setEnabled(false);
-    }
+    // This used to disable the spinner around setValue, because a valueChanged
+    // handler could run a long blocking operation and the spinner's mouse-release
+    // auto-repeat would then fire an extra increment. The only slider that
+    // disabled tracking was the time slider, and time steps no longer block --
+    // they are queued on the loader thread -- so the workaround is gone.
```
toloudis, line 460: what would go wrong if this were restored?

